PR #785, opened 2025-10-26
feat: Use varied narrative segment types during gameplay

### LINKED ISSUE #765: Use varied narrative segment types during gameplay
## Plain Language Summary
Right now all narrative segments show up labeled as "scene" even though the system supports dialogue, action, and transition types. Would be good to see varied segment types during actual gameplay.

## Current Feature
The narrative generation system has segment types defined (scene, dialogue, action, transition, ending) but only ever generates "scene" type segments. Looking at the code, `NarrativeController` calls `generateSegment()` without specifying a segment type, so it defaults to scene every time.

## Domain
- [x] Narrative Engine

## Enhancement Description
The system should intelligently pick segment types based on what's happening in the narrative. Something like:
- Use "dialogue" when characters are talking to each other
- Use "action" for combat, movement, or physical activities  
- Use "transition" for time passing or scene changes
- Use "scene" for everything else (setting descriptions, etc.)

This would make the narrative flow more interesting and give each segment type its own visual styling.

## Reason for Enhancement
Having varied segment types would make the game session interface more dynamic and easier to scan. Right now everything looks the same, which gets monotonous. The different types already have styling defined in `NarrativeDisplay.tsx` (different borders, colors, labels) but they never get used.

## Possible Implementation
Could add logic to `NarrativeController` that analyzes the narrative context or AI response to pick the right segment type. Maybe:

1. Check if the narrative contains dialogue markers (quotes, "said", etc.) → use "dialogue"
2. Check if it's describing action verbs or movement → use "action"  
3. Check if it mentions time passing or location changes → use "transition"
4. Default to "scene" for descriptive content

Or simpler: let the AI decide by including segment type in the JSON response format, then pass it through to `formatResponse()`.

## Alternatives Considered
Could just leave it as-is since it technically works, but that seems like a waste of the segment type system that's already built.

## Additional Context
Found this while investigating why visual tests only show "scene" segments. The type definitions in `narrative.types.ts:16` list all five types, and `NarrativeDisplay.tsx` has styling for each one, but the generation code at `narrativeGenerator.ts:407-409` defaults everything to scene.

Related files:
- `src/lib/ai/narrativeGenerator.ts:407-409` - where the default happens
- `src/types/narrative.types.ts:16` - type definitions  
- `src/components/Narrative/NarrativeDisplay.tsx` - segment styling
- `src/components/Narrative/NarrativeController.tsx:739` - where generateSegment gets called

jerseycheese:
This PR tackles two related problems with narrative generation: all segments were showing up as the same visual type even though the styling existed for variety, and the tone settings (particularly language complexity) weren't actually being enforced.

## What Changed

The work falls into two main areas that ended up complementing each other nicely.

### Varied Segment Types

The narrative display component already had styling for four different segment types - dialogue gets blue borders, action gets amber, transitions are gray, and scenes are the default white. But everything was hardcoded to 'scene' type in the generator, so that styling never activated.

The fix analyzes the narrative content after generation and infers the appropriate type based on simple pattern matching. If there are quotes or speaking verbs, it's dialogue. Combat or movement verbs make it action. Time or location changes become transitions. Everything else stays as a scene. This happens automatically without needing to change AI prompts - just looks at what got generated and categorizes it.

Added `segmentTypeInference.ts` with the detection logic, plus tests covering the different scenarios. The existing `NarrativeDisplay.tsx` component handles the rest since the styling was already there.

### Language Complexity Enforcement

The bigger issue was that tone settings weren't being respected. Setting language complexity to "Simple" would still generate complex, multi-clause sentences because nothing was validating the output.

Added a post-generation validation step that checks sentence length and vocabulary complexity. Each complexity level has thresholds:
- Simple aims for under 14 words per sentence with minimal long words
- Moderate allows up to 20 words per sentence  
- Advanced permits up to 28 words with richer vocabulary
- Literary has no limits

When content exceeds the thresholds, it gets automatically rewritten by the AI to match the target complexity. If the rewrite fails for some reason, the original content goes through but gets tagged for review.

This addresses feedback about overly complex language appearing even when simple language was requested. The example that prompted this was a 30+ word sentence full of complex vocabulary appearing despite "Simple" complexity settings. Now that gets caught and rewritten to shorter, clearer sentences.

## Files Modified

The segment type inference added three new files:
- `src/lib/utils/segmentTypeInference.ts` - pattern matching logic
- `src/lib/utils/segmentTypeInference.test.ts` - unit tests  
- `tests/visual/narrative-segment-types.spec.ts` - visual demonstration

Language complexity enforcement added:
- `src/lib/utils/languageComplexity.ts` - metrics calculation and validation
- `src/lib/utils/__tests__/languageComplexity.test.ts` - unit tests
- `src/lib/ai/__tests__/narrativeGenerator.toneSettings.test.ts` - integration tests

Both features integrate into `narrativeGenerator.ts` at three generation points (regular segments, initial scenes, and skill acknowledgments). The complexity enforcement runs after `formatResponse()` so it sees the final content.

Also updated `narrativeGenerator.loreContext.test.ts` to handle the potential second AI call when content needs rewriting.

## How It Works Together

The segment type inference happens first - it looks at the generated content and determines what kind of segment it is. Then the complexity enforcement checks if that content matches the world's tone settings. If not, it rewrites to simpler language while keeping the same segment type and story beats.

This means varied visual styling works correctly (dialogue looks different from action), and the language complexity actually matches what's configured for the world.

## Testing

All 1534 tests pass, which now includes:
- 14 tests for segment type detection  
- 5 tests for complexity validation
- Updated lore context tests to account for rewrite calls
- 378 lines of Playwright visual tests showing all segment types

Build and lint both clean.

## What This Doesn't Change

The AI prompts for narrative generation stay the same - no need to ask the AI to vary segment types since the inference handles it. The visual styling in `NarrativeDisplay.tsx` was already complete, just dormant. And the tone settings UI didn't change, this just makes the backend actually enforce what's configured.

Closes #765


jerseycheese:
## Visual Test Added

Added  to demonstrate all segment types with their distinct visual styling:

- 🟦 **Scene**: Default white background, gray text
- 🔵 **Dialogue**: Blue left border, italic text  
- 🟡 **Action**: Amber borders, bold text
- ⚪ **Transition**: Gray background, small italic text

The test creates isolated examples of each segment type and verifies they render with the correct styling.

**Note**: The existing  already includes all four segment types in the cyberpunk session (segments 1-4), so running  with the dev server will also demonstrate the feature in action.

To run the visual tests:
```bash
npm run dev  # Start dev server in one terminal
npx playwright test narrative-segment-types.spec.ts  # Run test in another
```

PR #1282, opened 2026-05-25
Add narrative text size control (#985)

### LINKED ISSUE #985: Add text density/size control for narrative display
## Plain Language Summary
A simple control that lets users change the size of the text in the story window. This means you can make the text smaller to fit more on the screen, or larger if you need it to be more readable.

## Description
Right now, the text size for the narrative is fixed at a specific size (it looks like `text-lg` in the code). While that works for a lot of cases, it's not ideal for everyone. Some users might want a "dense" mode to see more history without scrolling, while others might need larger text for accessibility or comfort.

This feature would add a simple control—maybe a toggle or a small settings dropdown near the narrative area—that lets the user pick their preferred density/size.

## User Story
As a player/reader, I want to be able to change the text size of the narrative so that I can either see more of the story at a glance (high density) or make it easier to read (low density).

## Acceptance Criteria
- [ ] Add a UI control to switch between text sizes (e.g., Small, Medium, Large).
- [ ] The "Medium" size should probably match the current default.
- [ ] The selection should be remembered (persisted) so the user doesn't have to set it every time.
- [ ] The narrative history should visually update immediately when the setting is changed.

## Implementation Notes
- The main narrative text styling seems to happen in `src/components/Narrative/NarrativeDisplay.tsx`, specifically around the line where `text-lg` is applied.
- We'll probably need a piece of state (maybe in `narrativeStore` or a new UI preferences store) to hold the current `textSize` setting.
- The control itself could live in `NarrativeController` or a header within the narrative display area.
- Need to make sure `FormattedNarrativeContent` inherits or receives this size so all parts of the text scale correctly.

jerseycheese:
## What this does

Lets readers change the narrative text size from the story view. There's a small Small / Medium / Large control in the narrative history header, and the pick sticks across sessions.

Medium maps to the existing per-theme size, so the default look is unchanged — you only see a difference if you opt into Small or Large.

Closes #985.

## How it works

- **`uiPreferencesStore`** — a small persisted Zustand store holding `narrativeTextSize`. Kept separate from session/navigation state so a reading-comfort setting survives across sessions and worlds.
- **`NarrativeTextSizeControl`** — segmented control wired to the store, with `role="group"` + `aria-pressed` on the active size.
- **CSS** — the per-theme `.text-narrative` font-size now multiplies by `--narrative-text-scale` (small 0.875, medium 1, large 1.25). The scale comes from `data-narrative-text-size` on the history container, which inherits down to every `.text-narrative` (history + live segment), so changes apply immediately. The history container became a flex column so the control header doesn't fight the scroll area.

## Acceptance criteria

- [x] UI control to switch text size (Small / Medium / Large)
- [x] Medium matches the current default (verified: ds1 base 13px at Medium)
- [x] Selection persisted (survives reload)
- [x] Narrative updates immediately on change (no reload)

## Testing

- Unit: store (default / set / persist) and control (renders options, reflects state, dispatches change) — Jest + RTL.
- Browser: drove the real game session on the dev route — Medium 13px, Large 16.25px, Small 11.375px, selection survived reload.
- Full Narrative + GameSession suites green (213 tests).

## Note on visual baselines

This adds a visible header to the narrative panel, so the game-session (and likely a few manuscript) visual baselines will diff by design. I'll adopt the regenerated baselines from the CI run rather than guess them locally.

### DIFF
--- a/src/components/Narrative/NarrativeHistory.tsx
+++ b/src/components/Narrative/NarrativeHistory.tsx
@@ -4,6 +4,8 @@ import { NarrativeDisplay } from './NarrativeDisplay';
 import { ScrollArea } from '@/components/ui/scroll-area';
 import { useBufferedNarrativeSegments } from './hooks/useBufferedNarrativeSegments';
 import { useTheme } from '@/lib/theme/ThemeProvider';
+import { NarrativeTextSizeControl } from './NarrativeTextSizeControl';
+import { useUIPreferencesStore } from '@/state/uiPreferencesStore';
 
 interface NarrativeHistoryProps {
   segments: NarrativeSegment[];
@@ -31,6 +33,9 @@ export const NarrativeHistory: React.FC<NarrativeHistoryProps> = ({
 
   const { theme } = useTheme();
   const { renderedSegments } = useBufferedNarrativeSegments(segments);
+  const narrativeTextSize = useUIPreferencesStore(
+    (state) => state.narrativeTextSize
+  );
 
   // Check if the viewport is near the bottom
   const getIsNearBottom = useCallback(() => {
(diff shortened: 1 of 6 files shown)


jerseycheese:
Shelving — reasoning on #985. Short version: browser zoom already covers the "make it bigger" need, and the only thing this adds is resizing the narrative text independently of the rest of the UI, which we decided isn't worth the surface area for a low-priority item. Branch stays put if we ever want the scoped behavior back.

PR #1527, opened 2026-07-12
refactor(design-system): unify onboarding tour onto --color-* tokens, remove legacy shadcn layer (#1474)

### LINKED ISSUE #1474: Unify drifted dual accent token systems (--primary vs --color-accent)
## Plain Language Summary
There are two parallel sets of theme color tokens left over from the Tailwind removal, and they've drifted apart. The onboarding tour reads the old set, so its accent color (a blue) no longer matches the app's real accent (indigo in DS1).

## Description
Found during the session-2 design audit. Post-Tailwind-removal (#1097), the legacy shadcn-style HSL tokens (`--primary`, `--background`, `--foreground`, `--card`, `--radius`, `--primary-foreground`, `--muted-foreground`) still live in `src/lib/theme/themes/ds{1,2,3}.css` alongside the canonical `--color-*` family. They aren't aliased to each other and their values have diverged.

Verified live (DS1):
- `--primary` = `221 83% 44%` (a blue, ≈ `rgb(19 78 207)`)
- `--color-accent` = `rgb(49 46 129)` (indigo)

So anything still reading `hsl(var(--primary))` renders a different accent than the rest of the app. The onboarding tour does exactly that — `src/lib/tutorial/tutorialConfig.ts` styles use `hsl(var(--primary))` / `--background` / `--foreground` / `--card` / `--radius` — so the tour's "Next" button is blue while the app's accent is indigo. shadcn-derived primitives (e.g. `dialog.tsx`) also still read the legacy family.

## Domain
- [x] State Management / Design System (theming tokens)

## Enhancement Description
Unify on one source of truth. Either alias the legacy tokens to the canonical ones per theme (`--primary: var(--color-accent)` etc.), or migrate the joyride config + remaining shadcn consumers to the `--color-*` family. Then the tour and dialogs track the real DS accent automatically.

## Reason for Enhancement
Two consequences: (1) a visible inconsistency — the tour is the first thing a new player sees and its accent is off; (2) a maintenance hazard — if someone deletes the "unused" `--primary` family, the tour and dialogs break, and meanwhile the two accents keep drifting independently.

## Possible Implementation
Map each legacy token to its `--color-*` equivalent in the three theme files (and the dark blocks), or sweep `hsl(var(--primary|background|foreground|card))` usages over to `--color-*`. Verify the tour Next button matches `--color-accent` in all three DS afterward.

## Alternatives Considered
Leave as-is — rejected; the drift is already visible and the dead token family is a footgun.

## Additional Context
This is technical-debt/architecture, distinct from the surface-level polish items filed alongside it.

jerseycheese:
## Description

Two parallel token systems were left in `ds3.css` after the Tailwind removal (#1097): the canonical `--color-*` family, and a legacy shadcn-style HSL family (`--primary`, `--background`, `--card`, `--radius`, …) read via `hsl(var(--token))`. #1474 flagged them for drifting apart.

Most of that drift is already gone — the DS3 collapse (#1526) deleted `ds1.css`/`ds2.css`, and in DS3 the two accents are nearly identical in light mode and only mildly apart in dark. What was left is the structural half: the legacy layer had exactly **one** runtime consumer, the onboarding tour, plus a single stray `var(--radius)`. That's the footgun — the layer looks unused, so someone deletes it and silently breaks the tour, while the two systems keep drifting.

This points the tour at `--color-*` directly and removes the legacy layer, so there's one token system. It's the item ADR-013 named as deferred ("removing the legacy shadcn HSL token layer").

- **`tutorialConfig.ts`** — the joyride tour's reads move onto `--color-*`: `hsl(var(--primary))` → `var(--color-accent)`, `--primary-foreground` → `--color-on-accent`, `--background` → `--color-canvas`, `--foreground`/`--card-foreground` → `--color-text-primary`, `--muted-foreground` → `--color-text-muted`. The tooltip/arrow background goes to the opaque `--color-overlay-surface-strong` (95/96%) rather than the 80% `--color-surface`, so it stays readable; radius → `--radius-md`.
- **`workshop.css`** — the one stray `var(--radius)` → `var(--radius-md)`.
- **`ds3.css`** — the `/* shadcn/ui HSL tokens */` block removed from both the light and dark blocks. Extended semantic HSL tokens (`--success*`/`--warning*`/`--info*`/`--ending-*`/`--lore-*`/`--alignment-*`) stay — they're still consumed.
- **Docs** — `design-tokens.md` Tier 3 and ADR-013's deferred list brought in line with the removal.

## Related Issue

Closes #1474

## Type of Change

- [x] Documentation update

## TDD Compliance

Token-reference refactor — no new logic, so no new tests (per house style, trivial styling isn't pinned with a spec). The tour's rendering is already covered by the existing `TutorialProvider`/`tutorialConfig` suites, which stay green.

- [ ] Tests written before implementation — N/A (no new behavior)
- [ ] All new code is tested — N/A (no new code paths; tour render covered by existing tests)
- [x] All tests pass locally

## Implementation Notes

The issue's original evidence is stale post-collapse: the dramatic blue-vs-indigo mismatch it describes was a DS1 artifact, and the "dialogs also read the legacy family" note no longer holds (the shadcn `ui/*` primitives were already re-skinned onto `--color-*`). The tour was the sole remaining consumer, which is why removing the whole layer is safe here.

Scope call: full removal of the legacy block rather than only re-pointing the accent tokens, so the codebase is left with one token system instead of a half-migrated two. That's what ADR-013 named.

Naming trap avoided: legacy `--accent` (a beige surface) is a different concept from `--color-accent` (the steel-blue brand) — it was unused, so it's just dropped, not mapped.

## Component Development

- [ ] Storybook stories created/updated — N/A (no component/markup change)
- [ ] Components developed in isolation first — N/A
- [x] Visual consistency verified (tour accent now matches the app accent in both modes — see below)

## Quality Checks

- [x] Linting passed (`lint` + `lint:css`)
- [x] Type checking passed
- [ ] Security audit passed — N/A (no security surface)
- [x] No unhandled promises

## Chrome DevTools MCP Verification Summary

Verified live against the running dev server (in-app browser, computed styles), with a probe element using the exact token references the joyride button now carries, portaled into the same `html[data-theme="ds3"]` subtree:

| | Light | Dark |
|---|---|---|
| Button background (`--color-accent`) | `rgb(91, 122, 140)` | `rgb(123, 160, 180)` |
| Button text (`--color-on-accent`) | `rgb(255, 255, 255)` | `rgb(23, 19, 16)` |
| Tooltip bg (`--color-overlay-surface-strong`) | `rgba(255, 252, 246, 0.95)` | `rgba(30, 26, 22, 0.96)` |

The deleted tokens (`--primary`, `--background`, `--card`, `--radius`) resolve empty at runtime in both modes — the legacy layer is gone, nothing re-defines it. The button values match the app's own accent buttons.

## Testing Instructions

1. `npm run type-check` / `npm run lint` / `npm run lint:css` — all clean.
2. `npm test` — 350 suites / 2356 tests pass.
3. Live: start the dev server, open the app; `getComputedStyle` on an element using `var(--color-accent)` returns `rgb(91 122 140)` (light) / `rgb(123 160 180)` (dark), and `--primary`/`--background`/`--card`/`--radius` return empty.

Note: the tour's dark-mode accent shifts slightly (from the drifted `rgb(131 157 170)` to the app's `rgb(123 160 180)`) and the tooltip goes from 100% to 95% opaque, so the **non-required** macOS-only Tutorial Visual baselines may need a refresh — not a merge gate.

## Checklist

- [ ] Comments added for complex logic — N/A (none warranted)
- [x] Documentation updated
- [x] No new warnings generated (removed one dead eslint-disable)
- [x] Accessibility considerations addressed (tour now uses the AA-audited `--color-accent`/`--color-on-accent` pair from #1480/#1483, in both modes)

### DIFF
--- a/public_docs/architecture/ADR-013-collapse-to-single-design-system-ds3.md
+++ b/public_docs/architecture/ADR-013-collapse-to-single-design-system-ds3.md
@@ -70,7 +70,10 @@ Also deferred, on purpose, and separately from each other:
 
 - **A bolder DS3** — deeper ink-blue accent, a louder dot grid, drafting marks, a real type scale, a brand-vs-product surface split. A design pass, not a code cleanup, and it changes token values again, which is exactly why this PR doesn't try to update `DESIGN.md`'s numbers yet (see the note added there).
 - **The `DESIGN.md` rewrite** that follows the bolder-DS3 pass, once there are real DS3 numbers worth documenting as canon instead of DS1's leftover values.
-- **Removing the legacy shadcn HSL token layer** (`--primary`, `--background`, etc. in [ds3.css](../../src/lib/theme/themes/ds3.css)) — the tour and some dialogs still read those tokens, so it stays until that's untangled.
+
+Since resolved:
+
+- **Removing the legacy shadcn HSL token layer** (`--primary`, `--background`, etc. in [ds3.css](../../src/lib/theme/themes/ds3.css)) — done in #1474. The onboarding tour was the last consumer once the shadcn `ui/*` primitives were re-skinned onto `--color-*`; it now reads the `--color-*` family too, and the legacy block was removed from ds3.css.
 
 ## Related Decisions
 
--- a/public_docs/design-system/design-tokens.md
+++ b/public_docs/design-system/design-tokens.md
@@ -4,7 +4,7 @@ type: design-system
 category: tokens
 tags: [design-tokens, colors, architecture, theming, dark-mode]
 created: 2025-01-09
-updated: 2025-06-15
+updated: 2026-07-12
 ---
 
 # Design Token System
@@ -119,35 +119,16 @@ These map intent and context onto the global foundation. Status feedback, storyt
 }
 ```
 
-### Tier 3: Component Tokens (shadcn-derived)
+### Tier 3: Extended Status Tokens (HSL)
 
-HSL-based tokens carried over from the shadcn/ui component foundation. They're consumed via
-`var()` in component CSS (typically wrapped in `hsl()`, e.g. `background: hsl(var(--primary))`).
-These predate the Tailwind removal, so despite the `bg-primary`-style naming there are no
-matching Tailwind utility classes — the values are reached through `var()` like every other token.
+A few status tokens are still stored as raw HSL channels and consumed via `hsl(var())`
+(e.g. `background: hsl(var(--success-background))`), because they carry background/border
+variants that predate the `--color-*` family. Everything else reaches color through the
+`--color-*` tokens (Tiers 1–2).
 
 ```css
 [data-theme="ds3"] {
-  --background: 36 24% 95%;
-  --foreground: 25 20% 14%;
-  --card: 38 100% 98%;
-  --card-foreground: 25 20% 14%;
-  --popover: 38 100% 98%;
-  --popover-foreground: 25 20% 14%;
-  --primary: 200 21% 45%;
-  --primary-foreground: 0 0% 100%;
-  --secondary: 25 14% 40%;
-  --secondary-foreground: 0 0% 100%;
-  --muted: 30 16% 92%;
-  --muted-foreground: 25 14% 46%;
-  --accent: 30 16% 92%;
-  --accent-foreground: 25 20% 14%;
-  --border: 29 16% 85%;
-  --input: 29 16% 85%;
-  --ring: 200 21% 45%;
-  --radius: 0.375rem;
-
-  /* Extended status tokens with background/border/muted variants */
+  /* Status tokens with background/border variants (HSL channels) */
   --success: 138 25% 40%;
   --success-background: 138 25% 90%;
   --success-border: 138 25% 75%;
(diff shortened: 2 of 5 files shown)


jerseycheese:
## Non-required CI failures — triaged, both pre-existing

Two non-required checks are red on this PR. Checked out `1ec0e5c6` (this PR's base, i.e. `develop` before this change) in the same worktree and reran each spec locally — both reproduce identically with none of this PR's changes present, so neither is caused by #1474:

**E2E Tests (1) — `characters-table-view.png`**: same 10195-pixel diff (ratio 0.01) on both this branch and `develop`'s tip. The diffed pixels are font-antialiasing on the Level/Type/Created columns and badge text, not a color-block shift — this page doesn't consume any token this PR touches. Stale baseline, unrelated to this change.

**Tutorial Visual Tests — `first-play-tour`**: flaky in CI (failed once, passed on retry) and passes clean locally once the dev server is warm. Matches the known cold-compile flake (#1520/#1521), not a regression.

**Tutorial Visual Tests — `character-creation-wizard-tour`**: this one's a real, deterministic failure (all 3 CI attempts + 2/2 local runs, same signature) — but it's already broken on `develop` itself. `develop`'s own CI run at `1ec0e5c6` shows the identical `waitForTooltip` failure. Root cause: at step 0, the tour tooltip renders with its bottom edge a few pixels past the viewport, so the helper's bounding-rect check never resolves true. Nothing to do with this PR's token changes (the helper only checks position, not color) — pre-existing breakage, and it doesn't look tracked under any open issue. Filing this as a separate follow-up rather than fixing it here, since it's out of scope for #1474.

Both jobs run `macos-latest` in CI, matching my local macOS run, so these are real reproductions — not the Linux-runner baseline gap.

Issue #1320, opened 2026-05-29
[Tracking] v1.0 release — visual finish + launch gate


jerseycheese:
## What this is

Tracking issue for the **next release to main (v1.0)**. The core MVP systems already shipped — world/character creation, AI narrative, journal, persistence, the DS migration, visual regression, onboarding.

**Re-scoped 2026-06-04.** 1.0 is a **free, public, single-player release**. The visual-finish work (Phases A/B below) is done. The original "launch gate" assumed launching as a *paid product* — but the app has no accounts, no backend persistence (local-first IndexedDB), and no billing, so monetization isn't a launch checklist, it's a separate product phase. That's been decoupled to the commercialization epic #495. What's left for 1.0 is the real free-launch gate: bring-your-own-API-key (so public visitors run on their own key, not the owner's), a simple landing page, and lightweight privacy/terms.

Scope first set 2026-05-29, re-scoped 2026-06-04. See `public_docs/development/mvp-roadmap.md` and the `v1.0` milestone.

## Dependency-ordered work

**Phase A — Visual blockers (Tailwind-removal fallout).** DONE.
- [x] #1316 — shared modals styled in production (established the pattern)
- [x] #1317 — remaining DS primitives (card/alert/tabs/table/checkbox/radio) styled in app
- [x] #1319 — data-table + CollapsibleSection rendered in the style guide

**Phase B — DS canon + structural differentiation.** DONE.
- [x] #1276 — showcase = canon (umbrella)
- [x] #1321 — Session phase: real game-session components in the showcase
- [x] #1165 — [EPIC] structural DS differentiation across surfaces
- [x] #1295 — character-side detail/edit content treatment + CharacterTable alignment
- [x] #1135 — About page real content + DS1/DS2/DS3 treatment

**Phase C — Free public launch gate** (re-scoped). The app reads as visually finished; this is what's actually needed to put it in front of strangers for free.
- [x] #891 — secure (encrypted) provider config + API key storage (PR #1371, merged to develop 2026-06-05)
- [x] #892 — provider config UI + presets + "no provider" blocking modal (PR #1371, merged to develop 2026-06-05)
- [x] #893 — route the existing API routes through the user's provider instead of the server env key (PR #1372, merged to develop 2026-06-05)
  - *Why: today all AI runs on one server-side `GEMINI_API_KEY` — every public visitor spends the owner's money. BYO-key flips the cost to the user, needs no accounts, fits local-first. This is the real 1.0 blocker.*
[...]

PR #1552, opened 2026-07-16
fix(settings): confirm before removing saved providers (#1537)

### LINKED ISSUE #1537: Confirm before removing saved providers
## Plain Language Summary
A saved BYO provider can be removed with one click from Settings -> Providers. That is risky because the provider disappears immediately, and if it was the only provider the app also clears the stored encryption key.

## Description
The Providers settings page treats `Remove` as an immediate destructive action. In the visual crawl, a saved "Visual QA Gemini" provider was visible on `/settings/providers`; clicking `Remove` took the page straight to the empty-provider state with no confirmation or undo step.

The code path matches what the crawl showed. `ProviderCard` calls `onDelete(provider.id)` directly, the settings page passes that straight through to `removeProvider`, and `removeProvider` clears the encryption key once the last provider is gone.

## Steps to Reproduce
1. Open `/settings/providers` with at least one saved provider.
2. Click `Remove` on the saved provider card.
3. Watch the provider list immediately switch to the empty state.

## Expected Behavior
Removing a saved provider should require confirmation first, especially when it is the active or only provider. Cancel should leave the provider untouched, and the encryption key should only be cleared after the user confirms removal of the last provider.

## Actual Behavior
The provider is removed immediately after the first click. When it is the last provider, the provider store also clears the encryption key after updating state.

## Screenshots
Captured in the fifth visual crawl:

- `/private/tmp/visual-crawl-20260713-fifth-pass/providers-saved-iphone-14-390x844-initial.png`
- `/private/tmp/visual-crawl-20260713-fifth-pass/providers-saved-iphone-14-390x844-after-remove-provider.png`
- `/private/tmp/visual-crawl-20260713-fifth-pass/providers-saved-desktop-1440x900-initial.png`
- `/private/tmp/visual-crawl-20260713-fifth-pass/providers-saved-desktop-1440x900-after-remove-provider.png`

## Environment
- Browser: Playwright Chromium
- OS: macOS
- Device: iPhone 14 viewport and desktop viewport

## Additional Context
Likely fix: route provider removal through the same shared confirmation pattern used for other destructive actions. Acceptance should cover the active-provider case and the last-provider case, with a focused component/page test that proves cancel does not call `removeProvider` and confirm does.

jerseycheese:
## Description
Remove on `/settings/providers` was a one-click destructive action — the provider disappeared immediately, and if it was the last one the store also cleared the encryption key with no warning. This routes provider removal through the existing `DeleteConfirmationDialog` (the same pattern `WorldListScreen` uses for worlds): the page holds the pending-removal id, `ProviderCard` stays dumb, cancel leaves everything untouched, and `removeProvider` only runs on confirm. The dialog copy adapts to the risky cases — removing your only provider warns that the stored encryption key gets cleared, and removing the active provider (with others saved) says another one takes over.

Part of the v1.0 launch-gate checklist (#1320).

## Related Issue
Closes #1537

## Type of Change
- [ ] Documentation update

## TDD Compliance
- [x] All new code is tested
- [x] All tests pass locally

The page had no spec at all before this; the new one covers the wiring the issue asked for (remove opens the dialog, cancel doesn't call `removeProvider`, confirm does).

## User Stories Addressed
As a player, I want a confirmation before a saved provider is removed, so one stray click can't delete my key — especially when it's the active or only provider.

## Flow Diagrams
N/A

## Component Development

N/A — no component changed. `ProviderCard` and `DeleteConfirmationDialog` are untouched; the fix is page-level wiring between two existing pieces.

## Implementation Notes
- `src/app/settings/providers/page.tsx` now tracks `providerToRemoveId`, and `onDelete` on the card sets it instead of calling `removeProvider` directly. The store method is unchanged, so the encryption key is still only cleared after the last provider goes — now behind an explicit confirm.
- Dialog message picks the sharpest warning: last provider > active provider > plain remove.
- One test-infra note: `@/`-alias CSS imports resolve through the jest alias mapper before the CSS stub rule can catch them, so the new spec stubs the page's stylesheet import. Pre-existing quirk — this page just never had a spec to trip over it.
- Per the Codex review finding: the page tracks an `isRemoving` flag and passes it through the dialog's existing `isDeleting` prop, with re-entry guards in both handlers. In-flight semantics: once confirmed, removal runs to completion — Cancel and Escape are inert while the await is running, both buttons are disabled, and the dialog closes when it settles. Double-clicking Confirm starts exactly one removal.

## Screenshots
N/A — the dialog is the existing `DeleteConfirmationDialog` in its destructive variant.

N/A

N/A

- [x] Linting passed
- [x] Type checking passed
- [ ] Security audit passed
- [x] No unhandled promises

Security audit not run separately — CI covers it. No `.css` touched, so no stylelint-relevant changes.

## Testing Instructions
1. `npm test -- src/app/settings/providers` — 4 specs: remove opens the dialog, cancel keeps the provider, confirming the last provider removes it and clears the key, removing the active provider warns about the switch.
2. Manually: open `/settings/providers` with a saved provider, click Remove — you get a confirmation instead of an instant delete. Cancel keeps the card; confirm shows the empty state.
3. With two providers, remove the active one — the dialog notes the other takes over, and it becomes active after confirm.

## Checklist

Docs: nothing to update — behavior now matches the existing destructive-action pattern. Accessibility rides on `ConfirmationDialog`'s existing focus handling (destructive variant focuses Cancel first).

### DIFF
--- a/src/app/settings/providers/__tests__/page.test.tsx
+++ b/src/app/settings/providers/__tests__/page.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { render, screen, waitFor, within } from '@testing-library/react';
+import { act, render, screen, waitFor, within } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 
 // The jest alias mapper resolves @/-form CSS imports before the CSS stub rule
@@ -46,6 +46,7 @@ jest.mock('@/components/DeleteConfirmationDialog/DeleteConfirmationDialog', () =
     description,
     itemName,
     confirmButtonText,
+    isDeleting,
   }: {
     isOpen: boolean;
     onClose: () => void;
@@ -54,14 +55,19 @@ jest.mock('@/components/DeleteConfirmationDialog/DeleteConfirmationDialog', () =
     description: string;
     itemName: string;
     confirmButtonText?: string;
+    isDeleting?: boolean;
   }) => {
     if (!isOpen) return null;
     return (
       <div role="dialog" aria-label={title}>
         <p>{description}</p>
         <p>{itemName}</p>
-        <button onClick={onClose}>Cancel</button>
-        <button onClick={onConfirm}>{confirmButtonText || 'Delete'}</button>
+        <button onClick={onClose} disabled={isDeleting}>
+          Cancel
+        </button>
+        <button onClick={onConfirm} disabled={isDeleting}>
+          {confirmButtonText || 'Delete'}
+        </button>
       </div>
     );
   },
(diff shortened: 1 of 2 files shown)


jerseycheese:
Good catch on the in-flight gap - the page now tracks an isRemoving flag and feeds it through the dialog's existing isDeleting prop, so both buttons disable during the await, the handlers guard against re-entry, and once you confirm the removal runs to completion (cancel/escape stay inert until it settles) - covered by two new specs in 24770a58.

Issue #1587, opened 2026-07-20
Assert layout and behavior invariants instead of leaning on pixel baselines alone


jerseycheese:
## Plain Language Summary
Visual testing here is entirely pixel comparison, so layout and behavior can break without any framed pixel changing. Three of them recently did. The fix is asserting on measured geometry and behavior, not on CSS properties.

## Current Feature
The visual suite is 42 `toHaveScreenshot` calls plus some presence checks (`toBeVisible`, `toBeEnabled`, `toHaveAttribute`) and a couple of numeric measurements. Nothing asserts a layout outcome.

## Domain
- [x] Other: Testing

## Enhancement Description

**Rewritten 2026-07-19.** The first version of this issue framed the gap as "zero `toHaveCSS` in the suite" and led with a `toHaveCSS` example. That's the weakest form of the idea and would have produced exactly the trivial tests this is meant to avoid. The gap is missing assertions on layout *outcomes*. Which API gets used is secondary, and `toHaveCSS` is usually the wrong one.

### Why asserting CSS properties mostly doesn't work

A computed-style assertion says a property resolved to a value. It doesn't say the value is doing anything.

#1582 is the proof. That bug is `float: right` declared on a child of a flex container. Floats don't apply to flex items, so the rule does nothing, but the declaration is right there and the computed value reads back as `right`. A `toHaveCSS('float', 'right')` check would have been green for the entire life of that bug. The same trap is available for `position: sticky` with no `top`, or inside an `overflow: hidden` ancestor, or against the wrong containing block.

The second problem is that it pins the technique. If the hero gap later comes from `gap` on a restructured parent instead of `margin-bottom`, an assertion on `margin-bottom` goes red on a page that looks correct. Then the assertion gets edited to match, which teaches you to treat that test as noise.

### What to assert instead

Measure the thing a player would notice.

```ts
// weak - passes even when sticky is inert
await expect(rail).toHaveCSS('position', 'sticky');

// real - scroll, then check the bar is still on screen
await main.evaluate(el => el.scrollTop = 600);
expect((await rail.boundingBox()).y).toBeGreaterThanOrEqual(0);
```

Same shape everywhere else. For the character-detail gap, measure the distance between the hero's bottom edge and the portrait's top edge rather than reading `margin-bottom`. For the world card, assert a bounded rendered height rather than `-webkit-line-clamp`. Each survives any implementation that produces the right result and fails on any that doesn't.

### Where a property check still earns a place

[...]

Issue #1574, opened 2026-07-20
[BUG] Play surface: narrative never auto-scrolls and the location bar scrolls away


jerseycheese:
## Plain Language Summary
New story text doesn't scroll into view when it generates, so you have to scroll down by hand every single turn. The location bar also scrolls off the top and never comes back.

## Description
Both of these were fixed and closed in #1436 (PR #1463, June 25) as findings F49 and F51. Both are broken again on current `develop`. The visual baselines were regenerated inside the DS-collapse PRs, so the suite adopted the broken states as correct, and the auto-scroll unit test stubs the geometry it needs, so nothing went red.

### Auto-scroll is a no-op (was F51)

The logic in `src/components/Narrative/NarrativeHistory.tsx` is intact and fairly thorough: a new-segment effect (`:56-66`), a delayed initial-load scroll (`:69-83`), a `ResizeObserver` re-anchor for streaming growth (`:263-276`), and keyboard nav for `End` / `PageDown` / arrow-snap (`:86-187`). All of it resolves `scrollViewportRef` to the Radix viewport at `:253-258`.

That viewport can't scroll. `.narrative-history-container` sets `height: 100%` (`src/styles/manuscript-session.css:107-113`), but nothing in the ancestor chain has a resolved height:

- `.manuscript-narrative-container` sets `width: 100%` only (`:102-104`)
- `.manuscript-main-content-inner` sets `max-width` and margins only (`:96-100`)
- `.manuscript-main-content` sets `grid-column` plus padding only (`:491-493`, `:2775-2777`)
- `.narrative-history-manager` has no CSS at all

`height: 100%` against an auto-height parent computes to `auto`, so the viewport grows to content height, `scrollHeight === clientHeight`, and every `scrollTo` call does nothing. The element that actually scrolls is `.manuscript-overlay-main` (`:86-94`), which no code in the repo ever touches. That's also why `End` and `PageDown` silently do nothing on the play surface.

It does work in the dev harnesses (`src/app/dev/narrative-system/page.tsx:285`), which constrain height differently, so isolated testing looks fine.

### Location bar isn't sticky (was F49)

PR #1463 scoped its fix to DS1 on purpose:

```css
/* F49: pin the location to the top of the DS1 scene-status rail so a crowded
   scene's character list can't scroll it out of view. DS2/DS3 render the
   location as an always-visible top bar, so the sticky is scoped to DS1 (the
   only theme whose rail scrolls). */
[...]

Issue #1543, opened 2026-07-14
[EPIC] Bolder DS3 redesign — real accent, dot grid, type scale, drafting marks, brand/product split


jerseycheese:
## Plain Language Summary
DS3 survived the DS1/DS2 collapse, but the redesign that was supposed to give it a real point of view never happened. It's still, in ADR-013's own words, "somewhat timid about its own concept." This epic is that redesign pass.

## Epic Description
ADR-013 collapsed Narraitor's three design systems down to one (DS3, "The Mechanical Manuscript") and deliberately kept the redesign work separate from the deletion work — different kinds of risk (deletion vs. design), and mixing them would have made the collapse PR impossible to review cleanly. DS3 as shipped is the same DS3 that existed alongside DS1/DS2: a subtle dot grid, an ink accent that doesn't lean hard into "drafting tool," no dedicated type scale, and `DESIGN.md` still documenting DS1's leftover numbers under an honesty-pass banner.

This is "Part B" of the original two-part handoff plan behind the collapse (Part A was the collapse itself — ADR-013 / PR #1526 — plus the legacy-token cleanup it deferred, #1474 / PR #1527, both shipped). Filing this now so it isn't lost before the v1.0 release scope is finalized.

## Domain
- [x] Other: Design System

## Goals
- Deeper ink-blue accent — the current accent doesn't lean hard into "drafting tool."
- Louder dot grid — currently a subtle background texture.
- Drafting marks — a visual motif not yet present (sketch lines, correction marks, manuscript annotations — whatever reads as "mechanical manuscript" rather than generic).
- A real type scale — DS3 currently borrows sizing rather than having its own considered scale.
- A brand-vs-product surface split — marketing surfaces (`/welcome`, `/about`, etc.) and the actual app currently share one visual language; give them deliberately distinct treatment.
- A real decision about imagery in list views. World and character cards currently reduce a generated 800x400 image to a ~144x72px thumbnail, because the DS3 card CSS discards the `Hero` structure and reuses only its `<img>` (`workshop.css:3677`, `:3711-3718`). That's a working dense-row idiom rather than a bug, but it makes generated art nearly invisible on the surface players browse most.

## User Stories
Split into sub-issues, one per facet:
- [x] #1617 — Drafting-marks motif (PR #1712)
- [x] #1621 — Accent depth + dot-grid intensity (PR #1676)
- [x] #1622 — A real type scale
- [x] #1623 — Brand vs. product surface split
[...]

Issue #1327, opened 2026-05-30
DS3 outcome callout as true page-gutter marginalia


jerseycheese:
## Plain Language Summary
On the DS3 "mechanical" theme, the decision-outcome note (e.g. CRITICAL SUCCESS) could sit out in the page margin beside the story column — true marginalia — the way the old design demo pinned it, instead of hugging the column's right edge.

## Current Feature
On DS3 desktop the outcome callout (`.choice-outcome-callout`) is right-aligned within the narrative column via `align-self: flex-end` (`src/styles/manuscript-session.css`, the DS3 `@media (width >= 1024px)` block). It reads as a margin annotation, but it stays inside the ~680px text column.

## Domain
- [x] Game Session UI / Design System

## Enhancement Description
Push the DS3 outcome callout out into the actual page gutter to the right of the narrative column — full marginalia, with the body text running full-width beside it. This is what the old hand-built DS3 demo did (`float: right; margin: 0 -260px 20px 24px` into the gutter).

## Reason for Enhancement
The real component can't reach the gutter today, and not for a small reason:
- The callout's parent `article.narrative-segment` is a flex column, so `float` is inert (that's why the current rule uses `align-self`).
- The narrative scroll container `.scroll-smooth` is `overflow-x: hidden / overflow-y: scroll`, which clips anything pushed past the column's right edge, and the outer `.manuscript-overlay-main` is `overflow: auto`. You can't set `overflow-x: visible` without breaking vertical scroll (the spec couples the axes).

So a faithful gutter treatment needs a structural change to *which element scrolls* — e.g. moving the callout out of the per-segment scroll flow into a positioned layer anchored to the shell's content area — not a CSS tweak. That's why it was split out here rather than bundled into #1276.

## Possible Implementation
Render the outcome callout (or a positioned copy) outside the narrative scroll container, absolutely placed into the right gutter and vertically aligned to its segment; or restructure the scroll boundary so the gutter lives inside the non-clipping region. Gate it to wide viewports only (the gutter doesn't exist below ~1248px) and keep the in-column `align-self` version as the fallback. Touches `ManuscriptSessionShell` / `NarrativeHistory` / `NarrativeDisplay` plus the session CSS, with real regression risk to scrolling — hence its own issue.

## Alternatives Considered
The shipped in-column right-align (PR #1324) — clean, no scroll risk, reads as a margin note. Good enough that the gutter version is a polish follow-up, not a blocker.

## Additional Context
[...]

PR #1592, opened 2026-07-20
Keep marginalia in the prose margin

### LINKED ISSUE #1582: [BUG] Marginalia definitions use an inert float inside the flex segment
## Plain Language Summary
Lore term definitions are supposed to sit in the margin beside the prose. They render as a left-aligned block stacked above it instead, pushing the story text down.

## Description
`src/components/Narrative/TermDefinition.tsx:50` renders `.manuscript-marginalia-definition` as a direct child of `.narrative-segment` (`src/components/Narrative/NarrativeDisplay.tsx:148-153`). Desktop styling at `src/styles/manuscript-session.css:3621-3634`:

```css
.manuscript-marginalia-definition {
  position: relative;
  float: right;
  clear: right;
  width: 200px;
  ...
}
```

`float` doesn't apply to flex items, and the parent is a flex container: `.narrative-segment` is `display: flex; flex-direction: column` at `src/app/globals.css:436-440`. So `float: right` and `clear: right` are both no-ops, and the definition renders as a 200px left-aligned block taking its own vertical slot in the column.

The rationale comment at `NarrativeDisplay.tsx:145-147` explains that a float only wraps content following it in source order, which is true for block flow and irrelevant here, so that comment is stale too.

Worth separating this from #1327. The `.choice-outcome-callout` in the same segment hit the same constraint and got an `align-self: flex-end` fallback (`manuscript-session.css:3135-3141`) with a candid comment about floats being inert. The marginalia definition never got that fallback, so it's a plain bug rather than a documented compromise.

## Steps to Reproduce
1. Play until a segment includes a recognised lore term with a definition.
2. Watch where the definition renders relative to the prose.

## Expected Behavior
Definition sits to the side of the prose the way the outcome callout does.

## Actual Behavior
Left-aligned 200px block stacked above the prose, pushing it down.

## Screenshots
N/A.

## Environment
- Branch: `develop` @ 36b9e04f
- Surface: `/worlds/[id]/play`, desktop widths

## Additional Context
From playtest triage against #1417. Related to #1436 F38-F41, which covered the DS1 version of this before DS1 was deleted.

Quickest fix is matching the callout with `align-self: flex-end`, which is one line and consistent with the neighbour. Setting `.narrative-segment { display: block }` would make both floats work properly, but that loses `gap: var(--space-3)` and needs a spacing audit across every segment, so it's the bigger swing.

True gutter marginalia for either element is tracked separately in #1327, and that one got harder after the collapse since DS3 now overrides the three-column stage to `1fr` (`manuscript-session.css:2761-2763` and `:3088-3099`) with `.manuscript-rail-spacer { display: none }`. There's no gutter to move into right now.

## Definition of Done
Per the convention in #1587, this closes with the invariant frozen as an assertion:

- a rendered marginalia definition sits to the right of the prose column's left edge, rather than starting flush with it

Asserting on position rather than on the specific CSS property matters here, because the whole bug is that the declared property (`float`) does nothing. A check for `float: right` would have passed throughout.

jerseycheese:
## Description

Lore term definitions were supposed to sit beside the prose. They rendered as a 200px block stacked above it, because `.manuscript-marginalia-definition` had `float: right` while its parent `.narrative-segment` is `display: flex; flex-direction: column` — and float is inert on a flex item. The declared property was right there in the cascade doing absolutely nothing.

#1582 suggested the cheap fix: `align-self: flex-end`, matching what `.choice-outcome-callout` already did. This PR went the other way and put the definition in the actual page gutter, because the reason the cheap fix existed was that true gutter placement was blocked — and that blocker turned out to be removable.

The blocker was the narrative scroll container clipping overflow-x. Widening `.narrative-history-container` past 100% when it contains a marginalia term (`:has()`), narrowing the segments column by the same amount, and clipping horizontally at `.manuscript-overlay-main` gives the note a real lane outside the prose without introducing a horizontal scrollbar. The note is then absolutely positioned at `calc(100% + gap)` with a vertical offset measured off the clicked term, clamped so it can't escape the scroller or overlap the next segment.

Heads up that this PR carries more than the one fix. Details under Implementation Notes — the short version is one related bug fix, one design change that the main fix makes necessary, and one small cleanup.

## Related Issue

Closes #1582

(Won't auto-close on merge to `develop` — needs closing by hand.)

## Type of Change

- [ ] Documentation update

## TDD Compliance

- [x] All new code is tested
- [x] All tests pass locally

## User Stories Addressed

As a player, when I tap a lore term in the story, I want the definition to appear beside the prose rather than shoving the paragraph I'm reading down the page.

## Flow Diagrams

N/A

## Component Development

The positioning depends on the play surface's scroll container and the real manuscript column width, neither of which exists in isolation — this is exactly the case where Storybook-green wouldn't prove anything. Verified in Playwright against the running app instead.

### What's in here beyond the marginalia fix

Worth being explicit, since a diff that quietly does four things is how the last round of DS3 regressions went unnoticed:

1. **The marginalia gutter fix** — the actual #1582 work. CSS `@media (width >= 960px)` block, plus the offset measurement in `NarrativeDisplay.tsx` and the clamping layout effect in `TermDefinition.tsx`.

2. **Possessive and word-boundary term matching** (`FormattedNarrativeContent.tsx`, `useTermDefinitions.ts`) — separate bug, same feature surface, found while testing this. Clicking "Arasaka's" didn't resolve a definition at all because the lookup key included the possessive suffix, and a term like "real" would match inside "reality" and wrap a fragment of a longer word in a button. Both are fixed with unit tests. Arguably its own PR; I kept it because it's the same component family, independently tested, and splitting it would cost a full CI cycle for a fix that's already green.

3. **`.choice-outcome-callout` goes back into the prose flow** — from `align-self: flex-end; width: 220px` to full width. This isn't unrelated churn: that 220px right-pull existed *only* as an approximation of gutter marginalia, and develop's own comment said so ("True page-gutter marginalia is blocked by the narrative scroll container's overflow-x clip — tracked as a separate follow-up"). This PR removes that blocker, so the workaround is obsolete. The call I made is that lore definitions earn the gutter and outcome callouts are narrative metadata that belongs inline. That's a design decision and it's reversible — flagging it rather than burying it. It also conflicts with #1327's premise, which assumed the callout would be the thing that moved into the gutter; I've commented there.

4. **Hiding the vestigial Radix scrollbar** — `.scroll-smooth` is the ScrollArea viewport inside `NarrativeHistory`. Since #1591 moved scroll ownership to `.manuscript-overlay-main`, that inner scrollbar is a leftover, and it renders on top of the new note lane. Note this does *not* hide the play surface's real scrollbar.

### Known limitation

The lane width is `min(12.5rem, calc(50vw - 21.25rem - ...))`, which works out to `min(200px, 50vw - 392px)`. That's 200px from about 1184px viewport up, but it narrows below that and bottoms out around 88px at the 960px breakpoint floor. An 88px note with 12px padding leaves ~64px of text, so `overflow-wrap: anywhere` is doing real work down there and it looks cramped. It's bounded and it doesn't break layout, but the 960–1100px band is not good. Filed as a follow-up rather than fixed here, since #1543 is about to rewrite this area anyway.

The `50vw` term also assumes the manuscript column is centred in the viewport, which holds post-collapse (the stage is `1fr` with `.manuscript-rail-spacer` hidden) but would break if a rail ever came back.

### Assertions, not baselines

Per #1587, the invariant is frozen as measured geometry in `tests/visual/manuscript-regression-assertions.spec.ts`. The load-bearing one is `definitionLeft >= proseRight + 16`.

I checked it isn't vacuous, since a geometric assertion that's true either way is just the old problem wearing a new hat. Reverting only the CSS fix and re-running:

- faithful revert (develop's rule, gated at `>= 1024px`, test viewport 1005px): fails, `definitionLeft: 0` — that's the mobile bottom sheet, so this catches the breakpoint change
- positioning-only variant (develop's float rules with the breakpoint lowered to 960px so the old layout actually applies): fails, `definitionLeft: 175` against a required `>= 859`

Both deterministic across repeat runs. The assertion catches the old positioning and the breakpoint independently. Bundle rebuild was confirmed between each variant, so neither result is a stale-bundle artifact.

The test is on the long side — it also guards that the prose column doesn't shift when the note opens, that segment height is unchanged, that nothing introduces horizontal overflow, and that the note stays inside the scroller. Those aren't padding; widening a container past 100% is exactly the change that breaks those things quietly.

## Screenshots

Three baselines regenerated, each in its own commit per #655: the choice-consequences shot (callout is now full width, so the two consequence badges sit side by side instead of stacked) and the two first-play tutorial steps.

Reviewed against `refs/remotes/origin/develop`. The branch was cut on top of #1574's work as recommended, so its commit list still shows those two commits — confirmed the effective diff re-applies none of it.

The worktree env-sync script that was originally bundled here has been split out to #1594.

N/A — verified through Playwright geometry measurement against the running dev server rather than DevTools.

- [x] Linting passed
- [x] Type checking passed
- [x] Security audit passed
- [x] No unhandled promises

## Testing Instructions

1. Play until a segment mentions a lore term you have a definition for, at a desktop width of 1200px or more.
2. Click the term. The definition should appear in the right margin, roughly level with the term you clicked, without the paragraph moving.
3. Scroll while it's open — the note should track its segment rather than drifting, and should stay inside the scroller instead of sliding under the header.
4. Click a possessive form ("Arasaka's"). It should resolve to the same definition rather than doing nothing.
5. Narrow to about 1000px. The note gets cramped (see Known limitation) but should stay on screen and not introduce a horizontal scrollbar.
6. Below 960px it should fall back to the mobile bottom sheet.

## Checklist

On that last one: pointer dismissal no longer yanks focus back to the trigger, and focus calls use `preventScroll`. With the note absolutely positioned, the old focus-restore was scrolling the surface out from under the reader.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jerseycheese:
Back to draft — the note's head gets clipped, and the term name is the part that disappears.

I ran this against a live dev server rather than only reading the diff, and the marginalia is unreadable for any lore description longer than a couple of lines. You get a fragment of the description starting mid-sentence, with no category label, no term name, and no type sub-label. At 1280px:

```
CASE={"defTop":138,"defBottom":377,"defH":239,"nameTop":176,
      "clipTop":236,"nameVisible":false,"hiddenPx":98}
```

Measured across widths and description lengths, all with the same seeded world:

| Width | Description | Note height | Clipped above the fold | Term name visible |
|---|---|---|---|---|
| 1440 | long | 239px | 98px | no |
| 1440 | short | 119px | 0 | yes |
| 1280 | long | 239px | 98px | no |
| 1280 | long, scrolled | 239px | 98px | no |
| 1005 | long | 401px | 260px | no |
| 1005 | short | 139px | 0 | yes |
| 960 | long | 530px | 389px | no |

So it isn't a narrow-viewport edge case — it reproduces at 1440px. The variable is note height against the segment's vertical slot, not viewport width. Short definitions are fine everywhere; realistic ones aren't.

### Root cause

`TermDefinition.tsx`, in the clamping layout effect:

```js
const clippingBottom = Math.min(
  scrollerRect.bottom,
  historyViewportRect?.bottom ?? scrollerRect.bottom,
  nextSegmentRect && nextSegmentRect.top > segmentRect.top
    ? nextSegmentRect.top - viewportInset
    : scrollerRect.bottom
);
const maximumTopOffset =
  clippingBottom - segmentRect.top - element.offsetHeight - viewportInset;
const adjustedTopOffset =
  maximumTopOffset < minimumTopOffset
    ? maximumTopOffset
    : Math.min(topOffsetPx, Math.max(minimumTopOffset, maximumTopOffset));
```

`clippingBottom` is capped at the *next segment's* top, so the note is being fitted into its own segment's slot rather than into the visible scroller area. A 239px note against a three-line segment doesn't fit, `maximumTopOffset` goes negative, and the `maximumTopOffset < minimumTopOffset` branch takes it — which pins the note's bottom and lets its top run off the top of `.manuscript-overlay-main`, where `overflow: hidden` cuts it.

The tie-break is backwards. When the note can't fit, the top is the part worth keeping: the name and category are what make it a definition rather than a floating sentence.

### The part I think is worth sitting with

The assertion in this PR passes at 1005px while 260px of the note is clipped, including the name.

It passes because `definitionTop` is only ever bounded from above:

```js
expect(geometry.definitionTop).toBeLessThanOrEqual(
  (geometry.clickedTermTop as number) + 8
);
```

Nothing bounds it from below against the visible region, so a note shoved up out of view satisfies every check in the test. `definitionScrollHeight <= definitionClientHeight` passes too, because the note's own box does contain its content — it's the ancestor doing the clipping.

That's the #1587 trap one level up. The convention was followed, the assertions are real geometry rather than property checks, and they still bless a broken result because the chosen invariant wasn't the load-bearing one. Whatever the fix ends up being, `nameTop >= clipTop` (or the note's top edge inside the scroller) should be in the test, since that's the invariant a player would actually notice.

### Not obviously worse-is-better than develop

Worth naming: on `develop` the note is in the wrong place — a 200px block stacked above the prose, pushing it down — but it's fully readable. This branch fixes the position and breaks the reading. For a player, that trade may well be a downgrade, which is why I didn't want to merge it and file a follow-up.

### Options

1. **Fix the clamp here.** Flip the tie-break to pin the top, and decide what the overflow does — extend past the segment (risks overlapping the next note), cap `max-height` with internal scroll, or let it run into the gutter below. That last bit is a real design call, not just a sign flip.
2. **Drop the gutter for now and take the one-line fix #1582 originally scoped** (`align-self: flex-end`, matching the callout). Small, safe, readable, and #1543 is about to rewrite this area anyway — gutter marginalia could land there with the type scale and column width settled in one pass.
3. Merge as-is and file the clip. I'd argue against, per the point above.

Everything else in the branch holds up: the possessive/word-boundary term matching is a genuine fix with tests, the `float` diagnosis is right, and the positioning assertion is non-vacuous — reverting only the CSS gives `definitionLeft: 175` against a required `>= 859`, deterministic across runs.

The env-sync script that was bundled here is now #1594. The lane-width issue is #1595, which is a subset of this and may fold in depending on which option we take.

PR #1634, opened 2026-07-27
Fix weekly code-health sweep findings

### LINKED ISSUE #1633: Weekly code-health sweep — 2026-07-27
Automated weekly code-health sweep, report-only — nothing was changed, no branch, no PR. Ran against a throwaway clone of `develop` @ `5b1954fd`: `knip` in both modes, the full `npm run lint`, `stylelint`, plus manual passes over the wizard layer, the API routes, and the store helpers. Everything below was checked against source before it went in.

Context for this one. The automated lanes are still mostly quiet — `knip` default reports nothing, `npm run lint` exits 0 with 148 warnings and no errors, `stylelint` exits 0. So `knip --production` and a hand pass over the character-creation wizard did the work. The top two findings both live in that wizard, and one of them is a live behavior bug: every step component reads its validation state from the wrong slot.

Ranked by value-to-effort. Twelve findings, nothing truncated.

### The character-creation wizard's validation plumbing is half-wired

- [ ] **Every wizard step reads its validation one slot off.** `src/components/CharacterCreationWizard/steps/BasicInfoStep.tsx:57` — **high**. The steps are 0-indexed everywhere they're written: `useCharacterCreationWizard.ts:57` comments them `// Step 0` through `// Step 4`, `stepValidators` is keyed `0`–`4` (`:82`), `useWizardState.ts:73` treats `currentStep === 0` as first, and `handleValidation` writes to `wizard.state.currentStep` (`CharacterCreationWizard.tsx:193`). But all four steps that render errors read 1-indexed: `BasicInfoStep` is `case 0` and reads `data.validation[1]`, `AttributesStep` is `case 1` and reads `[2]` (`:64`), `SkillsStep` is `case 2` and reads `[3]` (`:206`), `BackgroundStep` is `case 3` and reads `[4]` (`:64`).

  So each step renders the *next* step's error block. The path where it bites: `handleCreate` (`CharacterCreationWizard.tsx:196`) validates every step, and on the first failure calls `wizard.setValidation(i, validation)` and `wizard.goToStep(i)` with a 0-based `i`. The user gets sent back to the offending step, and that step reads a slot nobody wrote — so the errors that caused the bounce are invisible. Fix is four one-line reads, but it wants a test, because nothing currently covers it.

- [ ] **`onValidation` is required on five steps and three of them drop it, with a no-op blur handler documenting the gap.** `src/components/CharacterCreationWizard/steps/BasicInfoStep.tsx:26` — **high**. `CharacterCreationWizard.tsx:248` spreads `onValidation: handleValidation` into every step. Only `SkillsStep` ever calls it (`:166`). `BasicInfoStep:26`, `AttributesStep:39`, and `BackgroundStep:19` declare it as a *required* prop and don't even destructure it. `PortraitStep`'s validator is `alwaysValid`, so it's fine.

  The tell is `BasicInfoStep:53`: `const handleBlur = () => { // Validation will be triggered by parent component };` — an empty function wired to three `onBlur` handlers (`:92,106,129`). Since `handleValidation` is the only thing that sets `touched: true` mid-step, and `showErrors` at `:58` gates on `touched`, blurring a bad name on the first step marks nothing and shows nothing. Same shape as last week's inert-props class (#1611), except here the prop is load-bearing. Either wire the three steps up or drop the prop from their interfaces — the current state is a required callback that three of five consumers ignore.

### Dead code the tools found

- [ ] **Two exported dedup functions are dead, kept alive only by their own tests.** `src/state/loreStore.deduplication.ts:43` and `:207` — **medium**. `knip --production` reports both. `loreStore.actions.ts:37` imports exactly one thing from that module — `mergeFactsImpl` — and wires it at `:447`. `scanForDuplicatesImpl` and `checkDuplicateBeforeCreateImpl` have no non-test reference anywhere in `src/`; there's no `scanForDuplicates` or `checkDuplicateBeforeCreate` store action to reach them. That's roughly 90 of the module's 261 lines, plus most of the 393-line `loreStore.deduplication.test.ts`, which imports both by name (`:11,13`) and gives each its own `describe` block (`:28,316`). Deleting the pair takes the tests with it.

- [ ] **`fixSessionNarrativeCounts.ts` has one reference in the repo and it's the test for `fixSessionNarrativeCounts.ts`.** `src/lib/session/fixSessionNarrativeCounts.ts:12` — **medium**. `knip --production` flags the file. `fixExistingSessionNarrativeCounts` is exported, nothing imports it except `src/lib/session/__tests__/fixSessionNarrativeCounts.test.ts:1`. The module's own docblock explains it was moved off `sessionStore` to break an import cycle — the move happened, the call site never came back. It's the sole consumer of `repairSavedSessionNarrativeCounts`, which is declared on the store interface (`src/types/game.types.ts:85`), implemented at `sessionStore.ts:678`, and otherwise only ever appears as a `jest.fn()` in mocks. Both go together, or neither.

### Duplicated logic that already has a home

- [ ] **The two check-similarity routes are copy-paste twins.** `src/app/api/lore/check-similarity/route.ts:1` — **medium**. Its own header says "Mirrors the pattern from inventory similarity checking", and it does, literally: same body destructure, same `name1`/`name2` guard, same lowercase exact-match short-circuit returning `confidence: 1.0`, same `getDefaultConfig(resolveApiKey(request))` + `new GeminiClient(config)`, same JSON extraction, same `similar ?? false` / `confidence ?? 0.0` response shape, same 500 handler. Roughly 45 of ~100 lines each are identical; the only real difference is the prompt text and one optional `category` parameter. These are also the only two routes in `src/app/api/` that construct a `GeminiClient` inline. One shared handler taking a prompt builder would collapse both.

- [ ] **Five places hand-roll `/\{[\s\S]*\}/` when `extractJsonObject` is the house helper.** `src/lib/ai/parseJSON.ts:19` — **medium**. That module is the documented answer for pulling an object out of an LLM response, and six modules use it (`endingGenerator`, `storyCheckpointGenerator`, `narrativeGenerator.response.parse`, `toneSettingsGenerator`, `inventoryCategorizer`, `eventSignificanceValidator`). Five others reimplement it with the greedy regex: `src/lib/utils/aiResponseParser.ts:68`, `src/lib/ai/narrativeGenerator.metadata.ts:87`, `src/lib/generators/characterGenerator.ts:141`, `src/lib/services/worldCreationService.ts:96`, and both check-similarity routes (`:79` and `:66`). The regex and the helper's `indexOf('{')`/`lastIndexOf('}')` are behaviorally the same, so it's a safe swap — worth doing so parser fixes land in one place instead of eleven.

### Bundled small stuff, all mechanical

- [ ] **`showLabels` and `disabled` on `PointPoolManager` are declared and never read.** `src/components/shared/PointPoolManager/PointPoolManager.tsx:27` — **low**. Both are in `PointPoolManagerProps` (`:27,28`), neither is destructured at `:31`, neither appears again in the file. Same class as the `LoadingState` props #1611 cleared. Note `WorldTypeSelector` and `SkillRangeEditor` both implement a real `showLabels`, so the name isn't a dead concept — this one component just never wired it.

- [ ] **`layout` on `WorldTypeSelector` is declared and never read.** `src/components/shared/WorldTypeSelector/WorldTypeSelector.tsx:15` — **low**. `layout?: 'vertical' | 'horizontal'` appears exactly once in the file. The `size` prop on the same component was deleted in #1611; this is its sibling that got missed.

- [ ] **A whole reduce computes `totalCapacity` and nothing uses it.** `src/components/CharacterCreationWizard/utils/validation.ts:98` — **low**. Four lines walking every selected skill to sum available headroom, then never referenced. Lint already warns on it. `totalAllocated` right above it is live (`:117`), so it's just the one loop.

- [ ] **`errorStyles.input` is three keys holding the same string, two of them unused.** `src/styles/errorStyles.ts:15` — **low**. `border`, `focus`, and `combined` all equal `'input-error'`. Only `combined` has callers (`FormComponents.tsx:92,136`). The nesting is left over from the Tailwind era when the three were genuinely different utility strings; now it's structure implying a distinction that doesn't exist.

- [ ] **The `/app/dev` redirects in `next.config.ts` point at a 404 in production.** `next.config.ts:93` — **low**. Two `permanent: true` redirects map `/app/dev` and `/app/dev/:path*` to `/dev/*`. But `src/app/dev/layout.tsx:12` calls `notFound()` when `NODE_ENV === 'production'`, so in prod these permanently redirect users to a 404 — and permanent redirects get cached by browsers. They're leftovers from an app-router restructure that predates the dev-route guard.

- [ ] **Useless single-child fragment.** `src/components/GameSession/EndingScreen.tsx:293` — **low**. Reported by `react/jsx-no-useless-fragment`; one of the few lint warnings outside `tests/visual/`.

### Checked and clean — recorded so nobody re-chases them

- `knip` default mode reports nothing but the five expected config hints, which `knip.json` explicitly documents as intentional. `knip --production` is the lane that earned its keep this week — everything in the dead-code section came from it.
- `npm run lint` exits 0. 148 warnings, no errors, and 121 of them are `no-explicit-any` in `tests/visual/`, deliberately warn-level. Outside that, 13 are unused `eslint-disable` directives in `scripts/`, and the rest are the findings above.
- `npx stylelint "**/*.css"` exits 0 with no output.
- No skipped or todo tests anywhere, and one `TODO` comment in all of `src/` (a Storybook note at `src/stories/06-patterns/ui-patterns/GuidedFirstTimeExperience.stories.tsx:5`).
- Every duplicate-basename pair under `src/` except two is a `__mocks__` twin, which is the intended convention. The exceptions are `BasicInfoStep.tsx` (two different wizards, genuinely separate) and `useWizardState.ts`, which is already tracked as #1612.
- The `Logger` class and the `logger` singleton coexisting isn't a finding — `logger.ts:161` exports a default-context instance on purpose, and 89 files want a context string while 109 don't.

jerseycheese:
## Description
Clears the weekly code-health sweep from July 27. The main user-facing fix is the character-creation wizard validation path: every step now reads the same zero-based validation slot that the wizard writes, and the step validation callbacks are no longer dead props.

This also removes test-only dead code, deduplicates the two AI similarity route handlers, routes repeated LLM JSON-object extraction through the shared parser helper, and cleans the small mechanical findings from the sweep.

## Related Issue
Closes #1633

## Type of Change
- [ ] Documentation update

## TDD Compliance
- [x] All new code is tested
- [x] All tests pass locally

## User Stories Addressed
- Issue #1633: Weekly code-health sweep findings for the character-creation wizard, dead code, duplicated parsing/route logic, and small lint cleanup.

## Flow Diagrams
Not applicable.

## Component Development
- [x] Storybook stories created/updated

## Implementation Notes
- Character-creation steps now render validation from slots `0` through `3`, matching the wizard state and validators.
- Removed duplicate-scan/create exports and the narrative-count repair helper because they had no production callers.
- Shared similarity request handling now lives in `src/app/api/_shared/similarityCheck.ts`; inventory and lore routes keep their route-specific prompt text.
- `extractJsonObject` now handles the repeated LLM JSON-object extraction cases while existing caller-specific cleanup stays in place.

## Screenshots
Not applicable.

Self-reviewed the diff after local verification. The one stale issue note was `SkillsStep` UI `totalCapacity`: that value is live now, so only the actually unused validation-side reduce was removed.

Not run; this is code-health and unit/build verified.

- [x] Linting passed
- [x] Type checking passed
- [ ] Security audit passed
- [x] No unhandled promises

## Testing Instructions
1. `npm test -- src/components/CharacterCreationWizard/steps/__tests__/CharacterCreationStep.validation.test.tsx src/components/CharacterCreationWizard/steps/__tests__/BasicInfoStep.physicalDescription.test.tsx src/components/CharacterCreationWizard/steps/__tests__/SkillsStep.allocation.test.tsx src/app/api/lore/check-similarity/__tests__/route.test.ts src/state/__tests__/loreStore.deduplication.test.ts src/lib/lore/__tests__/fuzzyMatcher.test.ts`
2. `npm run type-check`
3. `npm run lint`
4. `npm run knip`
5. `npm run knip:production`
6. `npm run build:app`

## Checklist
- [ ] Comments added for complex logic


jerseycheese:
## Review

Reviewed against a fresh clone of `codex/issue-1633` @ `f4134e23`. Everything the PR body claims runs green — I re-ran all of it:

| check | result |
|---|---|
| `npm test` | exit 0 — 361 suites, 2380 tests |
| `npm run lint` | exit 0 — 146 warnings (was 148), 0 errors |
| `npm run type-check` | exit 0 |
| `npx knip` | exit 0 — only the 5 documented config hints |
| `npx knip --production` | exit 0 — completely clean (was 1 file + 2 exports) |
| `npm run build:app` | exit 0 |

### What this got right

The off-by-one is fixed correctly in all four steps, and the new test genuinely fails against the old code rather than just decorating the fix.

The dead-code removal is the standout. The sweep flagged `checkDuplicateBeforeCreateImpl`; this PR traced the whole cascade — that function was the only caller of `checkFactSimilarity`, which was the only caller of `checkLoreSimilarityClient`, which was the only user of the `SimilarityResult` type and `HIGH_CONFIDENCE_THRESHOLD`. All of it went, plus the orphaned tests. Same discipline on `repairSavedSessionNarrativeCounts`: down through `game.types.ts` and three separate mocks. That's more than the sweep asked for and it's right.

It also caught something the sweep got wrong. #1633 called `WorldTypeSelectorProps.layout` "declared and never read" — true of the component, but it missed that `src/app/worlds/page.tsx:225` was actually passing `layout="vertical"`, and that there was a whole `HorizontalLayout` Storybook story demoing a prop that never did anything. This PR removed the call site and the story. The original finding understated it.

### Issues

**1. The new blur validation disagrees with the wizard's own validator, and both write to the same slot.** This is the one worth blocking on.

`handleBlur` in `src/components/CharacterCreationWizard/steps/BasicInfoStep.tsx:55` now calls `validateCharacterName(name, worldId)`, which delegates to `validateName()` with its *defaults* — `minLength: 3`. But `stepValidators[0]` in `src/hooks/useCharacterCreationWizard.ts:84` uses `minLength(2)`. Ran both paths side by side:

```
name="Al"  blur → {valid:false, errors:["Name must be at least 3 characters"]}
           step → {valid:true,  errors:[]}
```

A user types a valid two-character name, blurs, and gets a red error. Hitting Next runs the wizard's validator, which says fine and advances. The copy diverges too — `"Name is required"` vs `"Character name is required"` — and the new test at `CharacterCreationStep.validation.test.tsx:76` asserts `"Character name is required"`, a string the new blur path never produces. It only passes because the test hand-feeds the validation object.

Root fix: pass `stepValidators[currentStep]` down from the wizard instead of re-importing individual validators into each step. Right now there are two sources of truth for step 0, and they disagree.

**2. The lore similarity route is now dead, and this PR refactored it anyway.** Deleting `checkLoreSimilarityClient` removed the only caller of `/api/lore/check-similarity`. The only remaining references are its own test file and the Storybook MSW handler at `.storybook/msw/handlers.ts:67`. So the shared-handler extraction — genuinely nice work — was spent partly on a route nothing calls. Either delete the route, its test, and the handler, or keep the client.

Worth knowing: `knip --production` can't catch this, because `route.ts` is a Next.js entry point. It'll stay invisible to the tooling. The inventory twin is still live via `checkItemSimilarityClient`, so the shared helper earns its keep either way.

**3. `BackgroundStep` validates on every keystroke.** `updateBackground` (`BackgroundStep.tsx:29`) runs `validateBackground` from each `onChange`, and that requires history >= 50 characters. So typing the first character sets `touched: true` and paints the error. `BasicInfoStep` correctly uses blur for the same job. Inconsistent, and the background step is now noisy in a way it wasn't before.

**4. `extractFencedJson` is stricter than the regex it replaced.** In `src/lib/utils/aiResponseParser.ts:62`, the old pattern was ` ```(?:json)? ` — optional language tag. The helper requires the `json` tag, so plain-fenced responses now fall through to `extractJsonObject` instead. It'll still parse (and picks up the truncation-repair path, arguably better), but it's an unflagged semantic change and `aiResponseParser.ts` has no test file at all.

**5. The test covers half the fix.** It asserts each step *displays* the right slot. `onValidation` is passed as a `jest.fn()` and never asserted — so the dead-prop half, which is where the divergence in #1 lives, has no coverage.

**6.** Nit: `src/components/GameSession/EndingScreen.tsx:294` — children left at the old indentation after the fragment came out.

### One correction to the PR body

The note about `totalCapacity` being a stale finding is right, and worth confirming rather than correcting: the live one is `SkillsStep.tsx:104`, the dead reduce was `validation.ts:98`, and only the dead one was removed. Good call.

---

Overall the mechanical work is solid — better than the sweep that produced it. The judgment slipped on #1 and #2. #1 is a real user-visible regression introduced by the fix, so that's the one to resolve before merge.

PR #1667, opened 2026-08-05
Rename the shared wizard hook to useWizardFlow

jerseycheese:
## Description

Two live hooks were both named `useWizardState` with incompatible shapes, so autocomplete on the name was a coin flip between two APIs — and `WorldCreationWizard` imports one of them alongside the other's directory in the same file.

The naming call: **the one under `components/shared/wizard` does more than hold state.** It persists to localStorage, runs a submit lifecycle (`onComplete`, error capture, clearing persisted state), and routes on cancel. So it becomes `useWizardFlow`, in `useWizardFlow.ts`. `@/hooks/useWizardState` keeps its name because in-memory step/data/validation state is exactly what it is — nothing about that name is wrong once the other one stops shadowing it.

Its exported types moved to match: `WizardState` -> `WizardFlowState`, `WizardConfig` -> `WizardFlowConfig`, `WizardHandlers` -> `WizardFlowHandlers`. None of the three had an importer outside the hook, so that rename is free.

Both hooks picked up a short header comment pointing at the other, so the next person picking one doesn't have to read both to tell them apart.

Consumers updated: `GuidedFirstTimeExperience.tsx`, `ai/ProviderWizard.tsx`, and the migration test (renamed to `useWizardFlow.migration.test.ts`).

Docs that named the hook were fixed rather than find-replaced. `GuidedFirstTimeExperience/README.md` claimed it shared a hook with the world creation wizard — it never did; those are the two different hooks. The shared-wizard technical guide's `### useWizardState` section is now `### Wizard state hooks` with a one-line "which one do I want" split. The shared/wizard README's example imported the hook from the barrel, which doesn't export it and never did, so that import now points at the real path.

## Related Issue

Closes #1612

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Refactoring (code improvements without changing functionality)
- [x] Documentation update
- [ ] Test addition or improvement

## TDD Compliance

Pure rename with no behavior change, so there's no new behavior to write a test for first. The existing coverage is the safety net — it renamed with the hook and still passes unchanged.

- [ ] Tests written before implementation — N/A, no behavior change
- [ ] All new code is tested — N/A, no new code
- [x] All tests pass locally
- [x] Test coverage maintained or improved

## User Stories Addressed

N/A — internal naming cleanup, no user-facing change.

## Flow Diagrams

[...]

PR #1668, opened 2026-08-05
Prune and re-scope the dependency-cruiser baseline (51 -> 17)

### LINKED ISSUE #1590: Prune and re-scope the dependency-cruiser baseline (51 entries, 31 reproduce)
## Plain Language Summary
The dependency-cruiser baseline suppresses 51 architectural rule violations, but only 31 still reproduce and about 14 of those are the rule failing to understand the code rather than real debt. The number has grown twice while nobody was looking at it.

## Current Feature
`.dependency-cruiser-known-violations.json`, the suppression baseline `npm run deps:validate` checks against.

## Domain
- [x] Other: Tooling / CI

## Enhancement Description

Three separate problems, and they want doing in this order.

**The baseline is stale.** 51 entries at `develop` @ 36b9e04f. Running `depcruise src --config` without the baseline reports 31 violations, so 20 entries no longer fire. Nearly all of the dead ones are `not-to-dev-dep` against `src/stories/**/*.stories.tsx`, which the rule's own `from.pathNot` already excludes at `.dependency-cruiser.cjs:214`. `npm run deps:baseline` regenerates it. Near-zero effort and it should happen before the number gets quoted again.

**Roughly 14 of the 31 that do fire are rule-scoping noise.** All 7 `no-orphans` hits are Jest manual mocks reached through bare `jest.mock('path')` calls, which dependency-cruiser can't see as an edge. Verified bare-call counts: `worldStore` 21, `characterStore` 19, `sessionStore` 17, `logger` 7, `geminiClient` 3, `ai/config` 2, `indexedDBAdapter` 1. All 7 `types-no-implementation-imports` hits are `typeof import()` type-only positions in `src/types/global.d.ts:16-22`. Scoping those two rules drops the real number to roughly 15.

**What's left is actual debt.** Around 8 `no-circular` and 7 `components-no-direct-lib-imports`, which is a burn-down rather than a config fix.

**It grows when unattended.** #1461 cited 46 in June. It's 51 now, last touched in `8fcdc3fe` on 2026-07-02. A suppression list nobody prunes stops being a baseline and becomes a place violations go to be forgotten.

## Reason for Enhancement
The first two steps are cheap and turn a meaningless 51 into a meaningful ~15, which makes the third step a tractable list instead of a wall. Right now the number can't be used to judge anything, since it's mostly stale entries and tooling blind spots.

Related: the `skott` circular-dependency ratchet already works this way and holds at 6, so there's a working pattern in the repo for a number that only goes down.

## Possible Implementation
1. `npm run deps:baseline` to drop the 20 stale entries.
2. Scope `no-orphans` to exclude `__mocks__`, and `types-no-implementation-imports` to allow `typeof import()` positions.
3. Re-baseline, then treat the remaining ~15 as a ratchet that can only shrink, matching how `.skott-baseline.json` is handled.

## Alternatives Considered
Leaving it. That's what produced the growth from 46 to 51 and three sweeps flagging it.

Deleting the baseline and letting `deps:validate` fail. Too blunt, since the ~15 real violations need actual refactoring and CI would stay red the whole time.

## Additional Context
Carried out of the weekly code-health sweeps #1461 and #1514, which are being closed as standing checklists per the precedent on #1569. Re-verified against `develop` on 2026-07-19 before re-filing.

Sibling re-files cover the wasted portrait AI calls and the dead store-code deletions.

jerseycheese:
## Description

The dependency-cruiser suppression list had drifted to 51 entries with nobody pruning it, so the number stopped being useful for judging anything. Three passes, in the order #1590 laid out.

**1. Re-baselined.** Dropped 20 stale entries, 51 -> 31. Nearly all of them were `not-to-dev-dep` against `src/stories/**/*.stories.tsx`, which the rule's own `from.pathNot` at `.dependency-cruiser.cjs:214` already excludes. They'd been sitting there suppressing violations that stopped firing a while ago.

**2. Scoped the two rules that were reporting tooling blind spots as debt.** 31 -> 17.

- All 7 `no-orphans` hits are Jest manual mocks under `__mocks__/`, reached through bare `jest.mock('path')` calls. Those aren't import statements, so dependency-cruiser can't see the edge and reads every mock as stranded. Added `(^|/)__mocks__/` to the rule's `from.pathNot`.
- All 7 `types-no-implementation-imports` hits are `typeof import()` positions in `src/types/global.d.ts:16-22`. They erase at compile time and pull in no implementation. Added `dependencyTypesNot: ['type-only', 'type-import']` to the rule's `to` clause. Both values, because dependency-cruiser tags these particular edges `type-import` rather than `type-only`, which is what the config's other rules use.

**3. Added `deps:check`, the other half of the ratchet.** `deps:validate` already blocks new violations landing. Nothing was stopping fixed ones from lingering as dead suppressions, which is exactly how the list got to 51. `scripts/check-deps-baseline.mjs` fails when a baseline entry no longer reproduces, so fixing a violation forces a re-baseline. Same idea as `check-circular-count.mjs` and `.skott-baseline.json`, one level more precise since it compares individual entries rather than a count. Wired into the Lint Check job right after `deps:validate`.

What's left is actual debt worth burning down: 8 `no-circular`, 7 `components-no-direct-lib-imports`, 2 test-only dev-dep imports.

## Related Issue

Closes #1590

## Type of Change

- [x] Documentation update

## TDD Compliance

N/A. This is CI tooling config with no product code touched, so there's no unit test to write first. Verification is by running the tooling itself, documented under Testing Instructions.

- [ ] All new code is tested
- [x] All tests pass locally
- [ ] Test coverage maintained or improved

## User Stories Addressed

N/A. Tooling and CI only, no user-facing behavior.

## Flow Diagrams

N/A.

## Component Development

N/A. No components touched.

## Implementation Notes

Both rule changes got a negative test before I trusted them, since a scoping change that quietly turns a rule off is worse than the noise it removes. Dropped a probe file into `src/types/` that does a real value import, and a stranded module under `src/lib/`, then ran the cruise:

```
warn no-orphans: src/lib/__tmpprobe/orphan.ts
error types-no-implementation-imports: src/types/__tmp_probe.ts -> src/state/worldStore.ts
```

Both still fire. The probes were removed.

The ratchet got the same treatment. Added a fake entry pointing at a file that doesn't exist and ran `deps:check`:

```
1 baseline entry no longer reproduces:
  dependency|x|src/nonexistent/ghost.ts -> src/lib/ai/presets.ts

Run `npm run deps:baseline` and commit the smaller file.
```

Exit code 1 on a stale entry, 0 on a clean baseline.

`public_docs/architecture/dependency-analysis.md` had stale counts throughout (51 violations, "630 lines", a "Type Violations (23 errors)" section that no longer describes anything). Updated the counts, rewrote the violation categories to match what's actually left, and documented both rule exemptions and the new command. `CLAUDE.md` gained the `deps:check` half of the gate and its failure mode.

## Screenshots

N/A. No UI changes.

N/A.

N/A.

## Quality Checks

- [x] Linting passed
- [x] Type checking passed
- [ ] Security audit passed
- [x] No unhandled promises

`npm run lint` exits 0 (144 pre-existing warnings, 0 errors). `npm run type-check` exits 0. `npm run knip` exits 0. Security audit not run, since no dependencies changed.

## Testing Instructions

```bash
npm run deps:validate:strict
```

Reports 17 violations: 8 `no-circular`, 7 `components-no-direct-lib-imports`, 2 `not-to-dev-dep`. Before this branch it reported 31.

```bash
npm run deps:validate
```

Passes with "17 known violations ignored".

```bash
npm run deps:check
```

Prints "Baseline holds at 17 suppressed violations, all still reproducing" and exits 0. To see it fail, hand-add an entry to `.dependency-cruiser-known-violations.json` pointing at a path that doesn't exist and run it again.

## Checklist

Accessibility is N/A. Nothing here renders.

### DIFF
--- a/.dependency-cruiser-known-violations.json
+++ b/.dependency-cruiser-known-violations.json
@@ -1,188 +1,71 @@
 [
   {
     "type": "dependency",
+    "from": "src/components/ai/ProviderPresets.tsx",
+    "to": "src/lib/ai/presets.ts",
+    "rule": {
+      "severity": "error",
+      "name": "components-no-direct-lib-imports"
+    }
+  },
+  {
+    "type": "dependency",
+    "from": "src/components/ai/ProviderWizard.tsx",
+    "to": "src/lib/ai/presets.ts",
+    "rule": {
+      "severity": "error",
+      "name": "components-no-direct-lib-imports"
+    }
+  },
+  {
+    "type": "dependency",
+    "from": "src/components/ai/ProviderWizard.tsx",
+    "to": "src/lib/ai/validateProviderClient.ts",
+    "rule": {
+      "severity": "error",
+      "name": "components-no-direct-lib-imports"
+    }
+  },
+  {
+    "type": "dependency",
+    "from": "src/components/forms/WorldImageForm.tsx",
+    "to": "src/lib/ai/worldImageGenerator.ts",
+    "rule": {
+      "severity": "error",
+      "name": "components-no-direct-lib-imports"
+    }
+  },
+  {
+    "type": "dependency",
+    "from": "src/components/WorldCreationWizard/steps/FinalizeStep.tsx",
+    "to": "src/lib/ai/worldImageGenerator.ts",
+    "rule": {
+      "severity": "error",
+      "name": "components-no-direct-lib-imports"
+    }
+  },
+  {
+    "type": "dependency",
+    "from": "src/components/WorldCreationWizard/WorldCreationWizard.tsx",
+    "to": "src/lib/ai/worldAnalyzerClient.ts",
+    "rule": {
+      "severity": "error",
+      "name": "components-no-direct-lib-imports"
+    }
+  },
+  {
+    "type": "dependency",
+    "from": "src/components/WorldCreationWizard/WorldCreationWizard.tsx",
+    "to": "src/lib/ai/worldImageGenerator.ts",
+    "rule": {
+      "severity": "error",
+      "name": "components-no-direct-lib-imports"
+    }
+  },
+  {
+    "type": "dependency",
     "from": "src/state/__tests__/loreStore.testHelpers.ts",
     "to": "node_modules/@testing-library/react/dist/@testing-library/react.esm.js",
-    "rule": {
-      "severity": "error",
-      "name": "not-to-dev-dep"
-    }
-  },
-  {
-    "type": "dependency",
-    "from": "src/stories/01-atoms/feedback/ErrorDisplay.stories.tsx",
-    "to": "node_modules/@storybook/test/dist/index.d.ts",
-    "rule": {
-      "severity": "error",
-      "name": "not-to-dev-dep"
-    }
-  },
-  {
-    "type": "dependency",
-    "from": "src/stories/01-atoms/feedback/SaveIndicator.stories.tsx",
-    "to": "node_modules/@storybook/addon-actions/dist/index.d.ts",
-    "rule": {
-      "severity": "error",
-      "name": "not-to-dev-dep"
-    }
-  },
-  {
-    "type": "dependency",
-    "from": "src/stories/01-atoms/forms/controls/RangeSlider.stories.tsx",
-    "to": "node_modules/@storybook/react/dist/index.d.ts",
-    "rule": {
-      "severity": "error",
-      "name": "not-to-dev-dep"
-    }
-  },
-  {
-    "type": "dependency",
-    "from": "src/stories/02-molecules/forms/editors/SkillRangeEditor.stories.tsx",
-    "to": "node_modules/@storybook/react/dist/index.d.ts",
-    "rule": {
-      "severity": "error",
-      "name": "not-to-dev-dep"
-    }
-  },
-  {
-    "type": "dependency",
-    "from": "src/stories/02-molecules/forms/managers/ToneSettingsForm.stories.tsx",
-    "to": "node_modules/@storybook/test/dist/index.d.ts",
-    "rule": {
-      "severity": "error",
-      "name": "not-to-dev-dep"
-    }
-  },
-  {
-    "type": "dependency",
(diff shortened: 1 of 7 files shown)


jerseycheese:
Tutorial Visual Tests is red here, and it is not from this branch.

The same job is red on `develop` itself at `54c7421c` (run [30955146931](https://github.com/jerseycheese/Narraitor/actions/runs/30955146931), push event, before this branch existed), and the same spec fails identically on every other open PR off that base: #1666, #1667, #1669.

The failing assertion is `tutorial-world-creation-finalize-20.png`, ~10005 pixels, consistent across all three retries and a full job re-run, so it is not the cold-compile flake from #1520. Comparing the artifact expected vs actual, the tour tooltip and everything below it sit about 2px lower than the baseline, plus a wider world-selector pill in the header. That is a real reflow, which is the same symptom #1671 is fixing.

Every other check passes, including the Lint Check job that now runs the new `deps:check` step. Nothing in this diff renders anything: dependency-cruiser config, an npm script, a Node CI script, a CI step, and docs.

Not re-baselining the tutorial snapshots here. That would mask the reflow rather than fix it, and it belongs to #1671.

jerseycheese:
Correction to the comment above: I pointed at #1671 as the fix. It is not, and it fails the same check itself. The tracked root cause is #1670.

`tests/visual/global.setup.ts:136` seeds the world image with a live `generatedAt` timestamp, which renders as `Generated: <today>`. The committed baseline holds the date it was recorded. That alone is only a handful of pixels, but `tutorial-world-creation-finalize-20` already burns ~10000 of its `maxDiffPixels: 10000` budget (`playwright.config.ts:65`) on anti-aliasing, so the date glyphs tip it to 10005 and it fails.

That also explains the develop run I cited: `54c7421c` passed on attempt 1 at 22:07 UTC on Aug 4 and failed on attempt 2 after the UTC date rolled over. Same commit, same code, different day.

Unrelated to this diff, and non-required. Leaving it to #1670.

Issue #1751, opened 2026-08-10
[BUG] Play surface is clipped on a 375px phone - the HUD row's min-content stretches every grid row past the viewport


jerseycheese:
## Plain Language Summary

On a phone, the right edge of the play screen is cut off. Skill badges on the suggested actions get sliced in half and there's no way to scroll over to see them.

## Description

`.manuscript-viewport-inner` is a grid with no explicit `grid-template-columns`, so its single implicit column gets an `auto` minimum: the widest row's min-content. The widest row is the HUD header, and at 375px it needs more than the viewport has.

Measured at 375x812:

| | width |
|---|---|
| viewport | 375px |
| `.manuscript-viewport-inner` content box | 351px |
| `.manuscript-hud-character-pill` (min-content) | 122px |
| `.manuscript-ds3-controls` (min-content, 7 icons) | 274px |
| header min-content (122 + 12 gap + 274) | **408px** |

That 408px propagates to the column, so `.manuscript-overlay-main` and `#manuscript-action-rail` both render 408px wide starting at x=12 — 45px past the right edge. `.manuscript-viewport-shell` has `overflow: hidden`, so the excess is clipped and unreachable.

There's a documented earlier attempt at this in the `@media (width < 640px)` block of [manuscript-session.css](../blob/develop/src/styles/manuscript-session.css), added as a #276 review follow-up ("The HUD icon row's auto-sized grid track overlaps the character pill's clickable area on narrow viewports once it grows past ~7 icons"). Two of its three rules are inert:

- `.manuscript-hud-reset-button` — no component uses this class. `ManuscriptFloatingHud` renders Reset with plain `manuscript-hud-icon-button`, so the Reset icon was meant to be hidden below 640px and never was.
- `.manuscript-tools-menu-item-mobile-only` — also unused anywhere in `src/`.

Only the shortcuts-button rule and the save-indicator rule actually fire. Seven icons still render at 375px.

## Steps to Reproduce

1. Open a play session at a 375px-wide viewport
2. Look at the right edge of the suggested actions

## Expected Behavior

The play surface fits the viewport. Choice text and skill-check badges are fully visible.

## Actual Behavior

Everything from x=363 rightward is clipped. `STEALTH · DC 4` renders as `STEALTH · DC` with the value cut off.

## Environment

- Browser: Chrome (driven over CDP)
- OS: macOS (Darwin 25.5.0)
- Server: `next dev` 15.5.19
[...]

Issue #276, opened 2025-05-03
[USER STORY] As a player with a keyboard, I want full keyboard navigation so I can play without a mouse (Complexity: Medium, Priority: Medium)


jerseycheese:
## User Story
As a player with a keyboard, I want full keyboard navigation so I can play without a mouse (Complexity: Medium, Priority: Medium)

## Acceptance Criteria

- 1. Narrative text is clearly displayed with appropriate formatting (paragraphs, emphasis, etc.)

## Technical Requirements

- - Responsive narrative display with:
- - Auto-scrolling to new content
- - Clear differentiation between narrative and dialogue

## Implementation Notes

- Use standard implementation approach for this feature
- Ensure compatibility with existing system architecture
- Write tests first following TDD approach

## Related Documentation
- [game-session.md](https://github.com/jerseycheese/narraitor/blob/develop/docs/requirements/ui/game-session.md)

## Estimated Complexity

- [ ] Small (1-2 days)
- [x] Medium (3-5 days)
- [ ] Large (1+ week)

## Priority

- [ ] High (MVP)
- [x] Medium (MVP Enhancement)
- [ ] Low (Nice to Have)
- [ ] Post-MVP

jerseycheese:
The infrastructure for keyboard control is mostly in place, but it's not wired up where it matters most.

What's already working:
- useKeyboardShortcuts hook handles modifier keys and prevents conflicts with input fields
- UI components have focus indicators built in (the ring that shows up when you tab through elements)
- Skip links let you jump to main content
- Modal dialogs trap focus automatically via Radix UI
- Navigation bar has Escape key support

What's missing:
- Game session has zero keyboard shortcuts - can't select choices, scroll narrative, or take any actions without a mouse
- No shortcut documentation anywhere for users to reference
- Tab order through the interface hasn't been verified to make sense
- Common actions like opening journal, viewing character sheet, or submitting custom choices need keyboard shortcuts

The foundation is solid but needs to be extended into the actual gameplay. The game session is where players spend most of their time, and that's exactly where keyboard control is currently missing.

jerseycheese:
[...]

PR #1752, opened 2026-08-10
Reserve the action rail so the story stops moving mid-turn

### LINKED ISSUE #1750: [BUG] Play surface jumps every turn - the action rail resizes the narrative viewport
## Plain Language Summary

Every time the story advances, the text you're reading physically jumps on screen. The buttons at the bottom change size as the story generates, and the story area shrinks and grows to compensate.

## Description

The play surface loses its footing on every single turn. It isn't a frame-rate problem (the page holds 118-119 fps the whole time), it's layout shift: the narrative viewport gets resized underneath the reader twice per turn.

`.manuscript-viewport-inner` lays the surface out as a three-row grid ([manuscript-session.css:46](../blob/develop/src/styles/manuscript-session.css#L46)):

```css
grid-template-rows: auto minmax(0, 1fr) auto;
```

Row 3 is the action rail, sized `auto` by its content. Row 2 is the narrative, sized `1fr`, so it absorbs whatever row 3 does. During a turn the rail's content swaps three times - idle choices, then `manuscript-choices-skeleton`, then the streaming rail, then a fresh set of choices whose text length differs from the last set. Each swap resizes row 3, and row 2 counter-resizes to match.

Measured heights across one turn:

| Element | Idle | Skeleton | Streaming |
|---|---|---|---|
| `#manuscript-action-rail` | 334px | 368px | 408px |
| `.manuscript-overlay-main` (narrative) | 552px | 517px | 461px |

The rail has a `max-height` ([:342-356](../blob/develop/src/styles/manuscript-session.css#L342)) but no `min-height`, and the skeleton only sets `min-width: 0` ([:436](../blob/develop/src/styles/manuscript-session.css#L436)), so nothing reserves the space.

There's a second, smaller contributor in the HUD that's independent of the rail: `manuscript-secondary-controls` ([ActiveGameSession.tsx:525](../blob/develop/src/components/GameSession/ActiveGameSession.tsx#L525)) collapses to zero mid-turn, which grows `manuscript-ds3-controls` from 46px to 63px and shoves `manuscript-hud-character-pill` sideways from x=17 to x=25.

## Steps to Reproduce

1. Start a game session with any world and character
2. Click a suggested action to advance the story
3. Watch the narrative text while the response generates

The jump is easiest to catch if you're mid-paragraph when you click.

## Expected Behavior

The narrative viewport keeps a stable height for the whole turn. Text stays where the reader left it while the next beat generates.

## Actual Behavior

The narrative area shrinks by ~91px as the turn starts and grows back as it finishes, moving the story text under the reader both times.

Measured with a `PerformanceObserver` on `layout-shift` over one isolated turn:

```
turn CLS:     0.1795   (Google "needs improvement" 0.1, "poor" 0.25)
shift count:  11
largest:      0.1058 and 0.0425
```

Cumulative across a few turns of normal play reached 0.5987. For contrast, frame timing is healthy: 118-119 fps sustained, 12 frames over 20ms in a 21s window, 15 DOM mutations per turn.

Pinning row 3 to a fixed height and replaying the same turn dropped CLS from 0.1795 to 0.0600, a 67% reduction, which points at the rail as the dominant cause. The 0.06 that survives is the HUD collapse described above.

## Screenshots

None - the problem is motion, so a still frame doesn't show it. The CLS numbers and the before/after A/B above stand in.

## Environment

- Browser: Chrome (driven over CDP)
- OS: macOS (Darwin 25.5.0)
- Device: Desktop, 120Hz display
- Server: `next dev` 15.5.19

## Estimated Model Power

- [ ] Light (mechanical, narrow, unambiguous - rename, copy tweak, dependency bump)
- [x] Standard (typical well-scoped work, follows an existing pattern - default tier)
- [ ] Advanced (cross-cutting or judgment-heavy - several viable approaches, real tradeoffs)
- [ ] Frontier (high-ambiguity, architecture-level or creative-judgment work)

## Additional Context

The obvious fix is reserving the rail row so choices, skeleton, and streaming all occupy the same box. The skeleton already carries fixed dimensions (`4.125rem` per action), so sizing it to match a real choice set and giving the rail a `min-height` would handle most of it. Another approach would be pinning row 3 outright in the grid template, though that risks clipping on short viewports, which is the thing the existing `max-height: min(50dvh, 26rem)` cap was added to avoid - worth checking against mobile before committing to it.

The `manuscript-secondary-controls` collapse is a separate small fix: reserve its space or keep it mounted rather than dropping it to zero.

Worth noting this was originally misdiagnosed as a streaming re-render problem. It isn't - instrumenting `Element.prototype.scrollTo` showed 7 calls per turn (5 from the `NarrativeHistory` ResizeObserver), and the streaming preview only updates about 7 times per turn. The per-token render path is fine.

jerseycheese:
## Description

The narrative viewport stopped moving mid-turn.

`.manuscript-viewport-inner` lays the play surface out as `auto minmax(0, 1fr) auto`. Row 3 is the action rail, sized by its content; row 2 is the narrative, so it absorbs whatever row 3 does. The rail's contents swap several times per turn, and each swap resized the story column under the reader.

Measured live at 886x863 with a `PerformanceObserver` on `layout-shift`, one turn:

| | rail | narrative |
|---|---|---|
| idle | 361px | 385px |
| choices skeleton | 376px | 372px |
| streaming | 435px | 312px |

Per-turn CLS **0.4335** across 10 shifts.

Four separate things were moving. All four are fixed here:

1. **The rail is a reserved box now.** `max-height: min(50dvh, 26rem)` became `height: min(50dvh, 24rem)`, with the internal scroll the cap already had. 24rem is the smallest reservation that fits a full turn — status line, context summary, three choices, and the custom-action row — without clipping the loading state. Narrow viewports were already sitting at that cap permanently, so nothing changes for them; desktop trades ~30px of reading space for a story column that never moves.
2. **The turn-status line has a slot.** "Continuing your story..." used to mount above the choices and push them down. It now renders into a slot that stays mounted whether or not a turn is in flight.
3. **The compact save indicator keeps its dot.** This is the HUD half of the issue, and the issue's diagnosis was off: `manuscript-secondary-controls` was a constant 17px, never collapsing mid-turn. The real cause was `SaveIndicator` swapping its status dot for a `LoadingState` spinner on every auto-save, which grew `.manuscript-ds3-controls` from 46px to 63px, pushed the whole narrative column down 17px, and slid `.manuscript-hud-character-pill` sideways. Compact saving now reads as a pulsing muted dot instead, matching the streaming indicator's vocabulary. The non-compact form still uses the spinner.
4. **The choices skeleton matches what it predicts.** Below 1024px the real choice prompt is hidden but the skeleton's placeholder wasn't, so the skeleton stood 40px taller than the choices replacing it. The skeleton now wears the same visibility class as the real selector.

Two smaller things the walkthrough turned up and this fixes alongside them: the context summary blinked out while the next decision generated (it holds the last one until a new one replaces it), and the sticky scene-status bar was 90% opaque, so story text stayed readable through it while scrolling.

Not fixed here, filed separately as #1751: the HUD's icon row plus character pill need 408px of min-content, which stretches every grid row past a 375px viewport and clips the right edge of the choice badges. Separate root cause, prior history in the #276 follow-up, and the fix is a product call about which controls survive on a phone.

## Related Issue

Closes #1750

## Type of Change

- [ ] Documentation update

## TDD Compliance

- [x] All new code is tested
- [x] All tests pass locally

Not test-first: the diagnosis needed live measurement in a real session before there was anything to write a test against. The regression test was written from the measured behavior and verified red-then-green against the unfixed code (0.1457 before, 0.0036 after) rather than being asserted into existence.

## User Stories Addressed

As a player reading a generated story, I want the text to stay where I left it while the next beat generates, so I don't lose my place mid-sentence every turn.

## Flow Diagrams

N/A

## Component Development

- [x] Components developed in isolation first

No new components and no new variants, so no new stories. `lint:ds-canon` passes (71 in-scope components, 0 gaps). Verified in the live app across light and dark, 375px and 1280px.

## Implementation Notes

- `24rem` is a measured number, not a round one: the skeleton needs 383px of content (16px status slot, 39px context summary, 218px of choices, 54px input row, plus gaps), and 24rem is 384px. Any smaller and the loading state clips.
- The rail's `overflow-y: auto` is unchanged, so a long choice set still scrolls internally the way the old cap intended.
- `.manuscript-characters-rail` gets its opacity by layering `--color-overlay-surface` over `--color-canvas` rather than a new token, so the bar keeps exactly the color it had while going opaque.
- The context summary is held in a ref rather than state on purpose — it's derived from props on every render and never needs to trigger one.

## Screenshots

N/A by nature: the bug is motion, so a still frame doesn't show it. The before/after is in the numbers below and in the changed `game-session` baseline, where the rail's reserved height and the narrative's new height are both visible.

## Visual Baseline Changes

Six, all from the rail's reserved height:

- `game-session.spec.ts-snapshots/game-session-chromium-darwin.png` — action rail is now a fixed 24rem box; narrative column shorter to match.
- `manuscript-layout.spec.ts-snapshots/manuscript-full-composition-chromium-darwin.png` — same.
- `manuscript-layout.spec.ts-snapshots/manuscript-hud-expanded-chromium-darwin.png` — same.
- `manuscript-breakpoints.spec.ts-snapshots/manuscript-drawer-inventory-desktop-chromium-darwin.png` — same.
- `tutorials/first-play-tour.spec.ts-snapshots/tutorial-first-play-step00-chromium-darwin.png` — the firstPlay tour's step 0 anchors on `narrative-display`, which the rail moved.
- `tutorials/first-play-tour.spec.ts-snapshots/tutorial-first-play-step01-chromium-darwin.png` — step 1 anchors on `player-choices`, inside the rail. CI only reported step 0 (the tour specs hard-assert, so the first diff hides the rest); regenerating locally caught both in one pass.

Regenerated locally with `--update-snapshots=changed`, then re-verified green. Local-vs-CI parity is evidenced by the same run: 52 of the 56 committed play-surface baselines matched local renders unchanged, so this Mac and the macOS runner agree on these specs. Same evidence on the tutorials side: the other eight tour specs' committed baselines, which came from the macOS runner, matched local renders unchanged.

Worth flagging: `manuscript-breakpoints.spec.ts` runs at `maxDiffPixels: 50000`, so only the one capture that crossed 53k pixels failed. Several of its other baselines are now stale-but-passing inside that tolerance. Left alone deliberately — tightening the threshold is its own change, and a blanket `--update-snapshots=all` on this spec set has corrupted baselines before.

N/A

Verified in a live session against this worktree's dev server with a real world, character, and Gemini generation — not mocks.

Baseline, before the fix, one turn: **CLS 0.4335**, 10 shifts, the two largest at 0.2438 (rail gaining `manuscript-action-rail-streaming` and growing 376 to 435px) and 0.1086 (HUD toolbar 46 to 63px on auto-save).

After, five consecutive real turns at 886x863: **0.0000, 0.0000, 0.0000, 0.0000, 0.0000** — zero `layout-shift` entries recorded. Cumulative across the run: 0.0000, against the issue's 0.5987. One mobile turn at 375x812 also read 0.0000. Rail held at 409px and the narrative at 354px for every phase of every turn.

The intermediate state is worth recording: with only the rail reserved, the turn read 0.0472 — under the bar but marginal, from the choice block sliding 14px as the context summary blinked out and back. Holding the summary took it to zero.

- [x] Linting passed
- [x] Type checking passed
- [ ] Security audit passed
- [x] No unhandled promises

`npm test` 393 suites / 2681 tests green, `type-check` clean, `lint` 0 errors, `lint:css` clean, `build` green, `knip` / `skott:check` / `deps:validate` / `deps:check` / `lint:ds-canon` all green. No security audit run — nothing in this change touches dependencies or input handling.

## Testing Instructions

1. `npm run dev`, start or resume a session, and click a suggested action. The story text should not move while the next beat generates.
2. `npx playwright test tests/visual/cls-measurement.spec.ts --project=chromium` — the new "taking a turn does not move the narrative viewport" case asserts CLS at or under 0.05 plus unchanged rail and narrative heights.
3. To measure it yourself, run a `PerformanceObserver` on `layout-shift` in a live session, zero it, click a choice, and read the total once the new choices land.

## Checklist

`ActiveGameSession.tsx` is 600 lines and predates this change; the diff doesn't grow it. On accessibility: dropping the compact spinner would have dropped its `role="status"`, so the status element carries `role="status"` with an `aria-label` of the same text the visible form would show — the saving state stays announced.


jerseycheese:
Parking this as a draft rather than closing it.

The reserve this PR adds is the thing a play-surface rethink takes back out. The plan now is one column, one scroll: the choices become the last beat of the transcript instead of a control panel under it, so there's no rail to reserve. Chasing a fixed-height box that holds unbounded content (recap, an optional ending block, 3-5 wrapping choice cards, the composer) is what's been producing the responsive edge cases and the stutter as pieces load and reload.

Worth keeping whatever happens, and worth pulling forward separately if this sits a while:

- SaveIndicator keeping compact manual saves disabled while a save is in flight
- Clearing the held context summary when a replacement decision legitimately has none
- The two re-recorded first-play tour baselines

Un-drafting is the move if the rework stalls. Everything here is still an improvement on develop.

jerseycheese:
Superseded by #1764, which fixed #1750 with a different approach (choices in the story flow instead of a reserved rail) and already merged.

PR #1772, opened 2026-08-12
Retire the token budget enforcement that never ran

jerseycheese:
## Description

Stacked on #1766 — review that one first; this targets its branch and will retarget to `develop` when it merges.

`NEXT_PUBLIC_ENABLE_TOKEN_BUDGET_MANAGER` shipped `false` in December 2025 and was never turned on. Twelve `applyBudget` truncation sites and a 75-line segment-dropping loop have been cold ever since, and the `budget?: RequestBudget` parameter threaded through roughly 34 signatures across six prompt-composition modules to serve them. That threading is the real cost — anyone touching prompt assembly had to understand a system that did nothing.

Before deleting it, I measured whether it was worth turning on instead. It isn't.

| Component | Cap | Realistic max | Can it bind? |
|---|---|---|---|
| `lore-context` | 800 | 544 | No — `getLoreContext` hard-limits to 20 facts, and nothing overrides it |
| `personalization` | 800 | 519 | No — name + attributes + skills + max 3 goals + max 2 hints, no growth term |
| `recent-narrative` | 1200 | 1620 | Only on three consecutive `critical`-weight beats |

`recent-narrative` is the one that looks unbounded and isn't. Every call site slices the window before the budget ever sees it:

| Call site | Window |
|---|---|
| `NarrativeController.tsx:622` | `slice(-3)` |
| `usePlayerChoices.ts:90` | `slice(-5)` |
| `useEndingDetection.ts:105` | `slice(-5)` |
| `useActiveGameSessionActions.ts:152` | `slice(-2)` |

Turn 5 and turn 500 send the same segments. Enforcement can't prevent bloat that never accumulates — all it could do is drop the oldest of three segments on exactly the turns where continuity matters most. That's a bug waiting to happen, not a safeguard.

Measurements come from the real Gemini tokenizer, with segment sizes taken from the app's own `narrativeLength.ts` guide (short = 3–5 sentences, medium = 1–2 paragraphs, long = 3–4 paragraphs).

**What stays.** The measurement half. One function records what a request's whole prompt weighed and reconciles that estimate against the provider's own token count, which is what the DevTools panel reads. That half is the reason the question above could be answered at all.

**Renames.** Two files stopped describing what they do, so `tokenBudgetManager` → `promptCalibration` and `narrativeGenerator.budget` → `narrativeGenerator.calibration`. `truncateToTokenLimit` went with them — its only callers were the truncation paths.

**Panel change.** The DevTools panel now shows one bar (whole prompt against the reference figure) plus the calibration summary, instead of eleven per-component bars. Per-component measurement is exactly what the threading paid for, so it goes with the threading. Priority labels and degradation hints go too — they described a resolver that no longer exists.

## Related Issue

N/A — follow-up to the audit in #1766, not a filed issue.

## Type of Change
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

Marked breaking because it removes a documented environment flag. Nothing observable changes for a player: the flag was off, so the deleted code never ran.

## TDD Compliance
- [x] All new code is tested
- [x] All tests pass locally

Not TDD — this is deletion. Tests that only exercised enforcement were removed rather than rewritten against nothing (`narrativeGenerator.budgets.test.ts`, the `tokenBudgetManager` suite, the lore-truncation case in `choiceGenerator.prompt.test.ts`, and the `truncateToTokenLimit` block in `tokenUtils.test.ts`). The calibration and panel suites were ported to the new shape.

## User Stories Addressed

N/A — internal simplification, no user-facing change.

## Flow Diagrams

N/A

## Component Development

The panel is gated behind dev-only flags, so the `DevToolsPanel` story renders but the bars don't. Its 6 unit tests cover the new rendering; there's no Storybook surface that would show the change.

## Implementation Notes

The rule throughout: prove the branch can't fire before deleting it. Every "can't bind" claim above came from running the real tokenizer against realistic inputs, not from reading the code.

`inventoryContextBuilder` keeps its own `tokenLimit` and `DEFAULT_TOKEN_LIMIT`. That bound is self-contained and still works — removing the budget-supplied override just means it falls back to its own default, which is what it did whenever the flag was off, i.e. always.

One thing worth flagging for review: `DEFAULT_TOTAL_BUDGET` (80,000) survives as the panel's reference denominator. It has never constrained anything, and it doesn't now, but as a "how big is this prompt, roughly" yardstick it's honest — the doc comment says so explicitly rather than implying a ceiling.

This also disposes of the second Codex finding on #1766 (`RequestBudget` installing component limits without respecting `totalBudget`). That constructor is gone.

## Screenshots

N/A — the only visual change is a dev-only panel that can't be reached without dev flags.

## Visual Baseline Changes

None. Doesn't touch `tests/visual/**-snapshots/**`, and no baseline covers `TokenBudgetPanel`.

N/A

N/A — no browser-reachable surface changed.

## Quality Checks
- [x] Linting passed
- [x] Type checking passed
- [ ] Security audit passed
- [x] No unhandled promises

Security audit not run — no dependency or auth surface changed.

## Testing Instructions

```
npm test              389 suites / 2641 tests passed   EXIT=0
npm run type-check                                     EXIT=0
npm run lint                                           EXIT=0
npm run lint:css                                       EXIT=0
npm run deps:validate                                  EXIT=0
npm run deps:check    baseline holds at 17
npm run knip                                           EXIT=0
```

To check by hand:

1. Play a turn and confirm the narrative reads the same — no prompt section should be shorter than before, since nothing was being truncated
2. Open the DevTools token-budget panel and confirm it shows the whole-prompt figure and, after a generation with a live key, the estimate-vs-actual accuracy

## Checklist

`public_docs/systems/narrative-generation.md` and `public_docs/features/ai-systems.md` both advertised the flag; both now describe what actually happens, which is that components are bounded where they're assembled and the system measures rather than trims.


jerseycheese:
Superseded by #1773. GitHub closed this one when `claude/audit-overengineering-06691d` was deleted on merge of #1766 — the base branch went away rather than retargeting. Same two commits, rebased onto `develop`, plus the font-size fix.

Issue #1749, opened 2026-08-10
Spike: can Gemini run through its own OpenAI-compatible endpoint?


jerseycheese:
> **Rewritten 2026-08-12.** Most of what this spike set out to test has since been answered from Google's own docs and their staff's forum replies, so the timebox shrank from "a few hours" to one narrow check. The original checklist and what each item resolved to is in the correction comment below.

## Plain Language Summary

Google ships an OpenAI-compatible endpoint for Gemini. Research has answered every question this spike was going to run except one: does its streaming output satisfy the consumer that PR #1717 built? That's a 30-minute check, and it's all that's left here.

## Current Feature

`src/utils/apiHelpers.ts` talks to Gemini's native REST API and shapes both the request and the streaming response by hand. #890 splits that into a provider-generic core plus per-provider adapters.

## Domain
- [x] Other: AI Service

## What's left to check

**Does the OpenAI-compatible endpoint's SSE stream satisfy the consumer from PR #1717?**

Gemini's native path is a chunked HTTP response; the compatibility path is OpenAI-shaped SSE with `data:` prefixes. `consumeGeminiStreamEvents` parses the former. Nothing in the docs says whether the latter drops cleanly into the same consumer or needs its own reader, and that's the one answer that changes how the adapter layer in #890 gets written.

Everything else on the original list is answered — see the comment below.

## Reason for Enhancement

Narrow as it now is, it's the difference between "the generic core reads one stream format" and "each adapter brings its own reader." Worth 30 minutes before #890 commits to a shape.

## Possible Implementation

Scratch script, not production code. Point it at `https://generativelanguage.googleapis.com/v1beta/openai/chat/completions` with `stream: true` and `stream_options: {include_usage: true}`, feed the chunks to the existing consumer, and see whether it survives. Write the answer into #890, then close.

## Alternatives Considered

Closing this outright and letting #890 discover it. Reasonable — but discovering it mid-refactor means rewriting the adapter boundary rather than choosing it, and the check is cheap.

## Estimated Model Power
- [x] Light (mechanical, narrow, unambiguous)

## Additional Context

Docs: https://ai.google.dev/gemini-api/docs/openai

Blocks #890. Part of the v1.2 milestone and #878.

Worth stating plainly since the original framing rested on it: even a clean yes here does **not** collapse this to one client. Image generation needs the native path regardless, and on the compatibility layer images live on a separate `/images/generations` endpoint rather than chat completions. So the saving is one text adapter, not one client.

[...]

Issue #895, opened 2025-11-20
Generic OpenAI-compatible provider with popular presets


jerseycheese:
## Plain Language Summary
Implement the generic OpenAI-compatible provider with presets for popular services (OpenAI, Deepseek, Mistral, OpenRouter, Together AI, Groq) so users can access dozens of AI models through one implementation.

## User Story
As a playtester, I want to use any OpenAI-compatible AI service (OpenAI, Deepseek, Mistral, etc.) so that I have maximum flexibility in choosing which models to use without needing provider-specific implementations.

## Acceptance Criteria
- [x] Generic OpenAI-compatible provider handles standard chat completions endpoint
- [x] Works with configurable endpoint URLs and models
- [x] Text generation uses standard `/v1/chat/completions` format
- [x] Content rating guidance injected in system prompts (universal baseline)
- [x] Error handling covers common failure modes (401, 403, 429, 5xx)
- [x] Presets added for: OpenAI, Deepseek, Mistral, OpenRouter, Together AI, Groq, Perplexity
- [x] Custom headers supported for services that need them
- [x] Works with all existing API routes through provider factory

## Technical Requirements
Implement `src/lib/ai/providers/openai-compatible/client.ts` with standard fetch:

```typescript
POST ${endpoint}
Headers:
  Authorization: Bearer ${apiKey}
  Content-Type: application/json
  ...customHeaders

Body:
{
  model: string,
  messages: [{ role: 'system' | 'user' | 'assistant', content: string }],
  temperature?: number,
  max_tokens?: number,
  top_p?: number
}
```

Add presets to `src/lib/ai/presets.ts`:
```typescript
const OPENAI_COMPATIBLE_PRESETS = [
  {
    name: 'OpenAI',
    type: 'openai-compatible',
    endpoint: 'https://api.openai.com/v1/chat/completions',
    models: ['gpt-4', 'gpt-4-turbo', 'gpt-3.5-turbo'],
    capabilities: { text: true, images: true, streaming: true },
    helpUrl: 'https://platform.openai.com/api-keys'
  },
  {
    name: 'Deepseek',
    type: 'openai-compatible',
    endpoint: 'https://api.deepseek.com/v1/chat/completions',
    models: ['deepseek-chat', 'deepseek-coder'],
    capabilities: { text: true, images: false, streaming: true },
    helpUrl: 'https://platform.deepseek.com/api_keys'
  },
  {
    name: 'Mistral AI',
    type: 'openai-compatible',
[...]

PR #1785, opened 2026-08-14
OpenAI-compatible provider abstraction layer (#890)

### LINKED ISSUE #890: OpenAI-compatible provider abstraction layer
## Plain Language Summary
Create a provider abstraction based on the OpenAI-compatible API standard, which unlocks support for dozens of LLM providers with a single implementation.

## User Story
As a developer, I need a provider abstraction built on OpenAI-compatible APIs so that the app can work with any provider that follows this standard (OpenAI, Deepseek, Mistral, OpenRouter, Together AI, Groq, and many others) without provider-specific code.

## Acceptance Criteria
- [ ] Provider interface defined based on OpenAI chat completions API format
- [ ] Provider interface implements Vercel AI SDK's `LanguageModelV1` specification
- [ ] Streaming support using Vercel AI SDK's Data Stream Protocol (multiplexes text, tool calls, metadata)
- [ ] Provider capabilities registry tracks per-model support (tool calling, image input, JSON mode)
- [ ] Message normalization middleware handles provider-specific quirks:
  - System prompt injection for models without dedicated system role
  - Message coalescing for Claude (requires alternating user/assistant messages)
- [ ] Safety abstraction maps content ratings to provider capabilities (best-effort)
- [ ] Provider configuration includes: endpoint URL, API key, model name, optional headers
- [ ] Generic OpenAI-compatible provider handles standard chat completions endpoint
- [ ] Provider factory instantiates correct provider based on configuration
- [ ] Existing API routes can use abstraction without knowing which provider is active
- [ ] Error handling covers common failure modes (invalid key, rate limit, timeout, moderation blocks)

## Technical Requirements
Create these files:
- `src/lib/ai/providers/types.ts` - Core interfaces (Provider, ProviderConfig, TextGenerationParams, SafetyParams, ProviderCapabilities)
- `src/lib/ai/providers/factory.ts` - Provider factory
- `src/lib/ai/providers/openai-compatible/client.ts` - Generic OpenAI-compatible implementation
- `src/lib/ai/providers/middleware/messageNormalizer.ts` - Message normalization logic
- `src/lib/ai/safety/SafetyService.ts` - Safety abstraction layer
- `src/lib/ai/safety/contentRatingMapper.ts` - Maps content ratings to provider-specific settings

Provider configuration schema:
```typescript
interface ProviderConfig {
  type: 'openai-compatible' | 'gemini' | 'claude' | 'ollama';
  name: string;  // User-friendly name
  endpoint: string;  // API endpoint URL
  apiKey?: string;  // Optional for local providers like Ollama
  model: string;  // Model identifier
  customHeaders?: Record<string, string>;  // Optional custom headers
  capabilities: ProviderCapabilities;
}

interface ProviderCapabilities {
  text: boolean;
  images: boolean;
  streaming: boolean;
  toolCalling: boolean;
  jsonMode: boolean;
  systemRole: boolean;  // Whether model supports dedicated system role
}
```

LanguageModelV1 interface implementation:
```typescript
// Must implement both methods from Vercel AI SDK
interface LanguageModelV1 {
  doGenerate(options: LanguageModelV1CallOptions): Promise<LanguageModelV1CallResult>;
  doStream(options: LanguageModelV1CallOptions): ReadableStream<LanguageModelV1StreamPart>;
}
```

OpenAI-compatible request format:
```typescript
POST ${endpoint}/v1/chat/completions
{
  model: string,
  messages: [{ role: 'user' | 'system' | 'assistant', content: string }],
  temperature?: number,
  max_tokens?: number,
  top_p?: number,
  stream?: boolean  // Enable streaming responses
}
```

Streaming protocol normalization:
- OpenAI: Server-Sent Events (SSE) with `data:` prefix
- Gemini: Chunked HTTP response
- Ollama: Line-delimited JSON
- Abstraction layer transcodes all to `LanguageModelV1StreamPart` format

## Implementation Considerations
The OpenAI chat completions API has become the de facto standard. Most providers either implement it directly or offer an OpenAI-compatible endpoint. This means one implementation covers:

- OpenAI (gpt-4, gpt-3.5-turbo, etc.)
- Deepseek (deepseek-chat, deepseek-coder)
- Mistral (mistral-large, mixtral-8x7b)
- OpenRouter (routes to any model)
- Together AI, Groq, Perplexity, and many others

**Vercel AI SDK Integration:**
Use the `LanguageModelV1` interface as the contract. This provides:
- Standardized prompt format (`LanguageModelV1Message[]`)
- Unified return types (`LanguageModelV1CallResult`)
- Built-in streaming support with Data Stream Protocol
- Type safety across all provider implementations

**Streaming Implementation:**
The Data Stream Protocol multiplexes different data types in a single stream:
- `0:"text"` - Text deltas
- `1:{"tool":"name"}` - Tool calls
- `2:{"key":"value"}` - Metadata

This prevents breaking when models switch from text to tool calls mid-generation.

**Message Normalization:**
Some providers have strict requirements:
- **System role**: If model doesn't support it, prepend system content to first user message
- **Message coalescing**: Claude requires alternating user/assistant messages. If user sends two consecutive messages, merge them before sending to API.

**Provider Capabilities:**
Don't assume all OpenAI-compatible endpoints support everything:
- Some models can't do tool calling (return descriptive error: "Model X doesn't support tools")
- Some can't handle images (fail fast with clear message)
- Track capabilities per model in the registry

The abstraction should be flexible about endpoint URLs. Some providers use `/v1/chat/completions`, others use `/chat/completions` or custom paths. Let users configure the full endpoint URL.

Safety/moderation varies wildly across providers. Some have moderation APIs (OpenAI), some block inline (Gemini via OpenAI-compatible endpoint), some just refuse in responses (most others). The safety layer should inject content rating guidance in system prompts as a universal baseline, then add provider-specific safety where available.

Don't over-engineer this. Start with the basic chat completions interface and add complexity only when needed for specific providers.

## Estimated Complexity
**Large** (8-12 hours) - This is foundational work that affects everything else

## Priority
**High** - Must be completed before any provider-specific implementations

## Domain
- [x] AI Service

## Definition of Done
- [ ] All interfaces defined and documented
- [ ] `LanguageModelV1` interface correctly implemented
- [ ] Streaming works with Data Stream Protocol
- [ ] Provider capabilities registry prevents unsupported operations
- [ ] Message normalization handles system role and coalescing
- [ ] Generic OpenAI-compatible provider works with standard endpoints
- [ ] Provider factory can instantiate providers based on configuration
- [ ] Safety service injects content rating guidance in system prompts
- [ ] Unit tests verify abstraction works with mocked responses
- [ ] Code follows existing patterns in `src/lib/ai/`

## Related Issues/Stories
Part of Epic #878 - Provider-Agnostic AI Support (MVP)

Blocks all other provider implementations

jerseycheese:
## Description

Splits `src/utils/apiHelpers.ts` into a provider-generic core plus per-provider adapters, and plumbs the player's full provider configuration through to the server so it can reach something other than Gemini.

Three things blocked this before:

1. `apiHelpers.ts` was 555 lines of Gemini — hardcoded REST URLs, hand-built `contents: [{parts: [{text}]}]` bodies, `candidates[0].content.parts[0].text` parsing — all mixed in with rate limiting and error responses.
2. `resolveApiKey` returned a bare string and `aiFetch` forwarded only a key, so `config.type`, `config.endpoint` and `config.model` never reached the server.
3. `createDefaultGeminiClient` and `clientFactory` branched on environment only, and every branch ended at Gemini.

The adapter boundary comes from the #1749 spike rather than guesswork. It replayed real network chunks from both Gemini endpoints and found the SSE framing identical (`200 text/event-stream`, `data:` prefixes) and only the payload inside each frame different — so the core keeps the decode loop, the partial-line buffer, the preview extraction, the monotonic-reveal guard and the event protocol, and each adapter supplies a frame extractor. The spike also found that compat frames run through the native consumer produce **zero deltas and a done event with `finishReason: "STOP"`**, indistinguishable downstream from a model that returned nothing. There is a test for exactly that case.

## Related Issue

Closes #890

## Type of Change

## TDD Compliance

- [x] All new code is tested
- [x] All tests pass locally

Not strict TDD. The behaviour being preserved already had tests, so those were the spec: the pre-existing `apiHelpers`, `resolveApiKey` and `resolveModel` suites pass untouched apart from two renamed entry points, which is the evidence that Gemini is unchanged. New tests were written alongside each new module.

## User Stories Addressed

> As a developer, I need a provider abstraction built on the OpenAI-compatible standard so the app can work with any provider that follows it without writing provider-specific code for each one.

Acceptance criteria, with what each landed as:

- [x] **apiHelpers split** — `src/lib/ai/providers/core/{request,streamConsumer}.ts` plus `gemini/adapter.ts` and `openai-compatible/adapter.ts`. Gemini's path behaves as before.
- [x] **Provider interface extends `AIClient`** — `OpenAICompatibleClient` implements the existing contract in `src/lib/ai/types.ts` beside `GeminiClient`.
- [x] **Factory dispatches on configuration** — `providers/factory.ts`; `createDefaultGeminiClient` now takes a resolved descriptor.
- [x] **`resolveApiKey` resolves a descriptor** — `resolveProvider()` returns type, endpoint, model and key. `resolveApiKey` and `resolveModel` are narrow views of it, so the model-validation rule has one implementation rather than two.
- [x] **`aiFetch` and `providerKeyHeader` forward type and endpoint** — new `x-provider-type` and `x-provider-endpoint` headers.
- [x] **Capabilities registry** — per-model streaming, images, JSON mode, dedicated system role, plus alternating-turns and native-safety-settings.
- [x] **Message normalization** — system-prompt injection for models with no system role (Gemma, Claude-family), coalescing for models that require alternating turns (Mistral-family).
- [x] **Content rating degrades honestly** — Gemini keeps `safetySettings`; providers without an equivalent get the rating as system-prompt guidance, and the config UI says plainly that it is guidance rather than an enforced filter.
- [x] **Privacy disclosure in the config UI** — see below.
- [x] **`ProviderConfig.model` is the model used** — `resolveProvider` carries it end to end, with a wider id pattern off Gemini so vendor-prefixed ids like `meta-llama/Llama-3.3-70B-Instruct-Turbo` survive.
- [x] **Routes work through the abstraction** — the narrative routes call `processAITextRequest` / `processAIStreamingTextRequest`; the generator routes forward a descriptor. No route names a provider.
- [x] **Error handling across providers** — invalid key (401/403), rate limit (429), timeout, and moderation blocks, including the `finish_reason: "error"` shape OpenRouter uses instead of `content_filter`.
- [x] **Gemini behaviour unchanged, SSE from #1717 intact** — the pre-existing suites are the proof.

## Component Development

- [x] Components developed in isolation first

`ProviderDisclosure` is a small text block inside the existing provider wizard and follows the surface's established classes; no new story was added. Its copy is covered by unit tests rather than a visual one.

## Implementation Notes

**Privacy, stated before the key gets pasted.** Google's free Gemini tier allows prompts and outputs to be used to improve their models and reviewed by human raters; their paid tiers do not. Someone writing personal creative fiction into this app should know that before they choose. Presets carry a `privacyNote`; a custom endpoint gets the honest version, which is that we have no way to know its terms.

**Content rating, without overselling either side.** On Gemini a world's rating goes out as a safety-filter setting. Everywhere else there is nothing to send it to, so it rides in the prompt as guidance. The disclosure doesn't oversell the Gemini side either: Google's own docs now put the default block threshold at "Off" for current models, and the non-adjustable core-harm layer is not reachable by any setting.

**Security.** This adds a URL that the *server* dereferences on a player's behalf, so:

- Routing headers are honoured only when the caller supplied their own key. On the env-key path the provider is always Gemini — otherwise any anonymous caller could name an endpoint and have the server post the deployment's key to it. That is key exfiltration, not just cost.
- Endpoints must be `https` on a public host. Loopback, RFC 1918, unique-local, and the `169.254` metadata range are refused.
- The Gemini adapter pins its own base URL and ignores `descriptor.endpoint`, so the path every default session uses has no attacker-reachable URL at all.
- `resolveApiKey` returns null when the active provider is not Gemini, rather than handing a player's OpenRouter key to Google. The image routes are Gemini-only by design, so "no key" is the honest answer there.

**Scope held deliberately.** The named presets other than Gemini stay `available: false` — verifying each of those is #895. The custom-endpoint form is enabled, since that is the generic door onto exactly what this abstraction provides.

**Two files named in the issue, deliberately left alone, rather than half-done:**

- `eventSignificanceValidator.ts` still constructs `GoogleGenAI` inline. Routing it through a client would drop the `temperature: 0.1` / `maxOutputTokens: 200` it deliberately uses for a yes/no classification, because `AIClient` has no per-call generation options. That wants its own change.
- `inventoryCategorizer.ts` stays on a bare Gemini key. On another provider it degrades to its existing deterministic fallback — the behaviour it already has when no key is configured.

## Visual Baseline Changes

None.

## Quality Checks

- [x] Linting passed (`npm run lint`, `npm run lint:css` — 0 errors)
- [x] Type checking passed (`npm run type-check`)
- [ ] Security audit passed
- [x] No unhandled promises

`npm audit` was not run; the dependency set is unchanged by this PR. `npm run deps:validate` and `npm run deps:check` both pass with the baseline unchanged at 17 — the one new boundary violation this introduced (a component reaching into `lib/ai`) was fixed with a colocated `use*` hook rather than baselined.

## Testing Instructions

**Automated**

```bash
npm test            # 2760 passed, 404 suites
npm run type-check
npm run lint && npm run lint:css
npm run deps:validate && npm run deps:check
npm run build       # needs NEXT_PUBLIC_SITE_URL set
```

**Gemini is unchanged** — the load-bearing check. Play a story turn with a Gemini key configured: narrative streams progressively as before, choices generate, the content rating still reaches Google as `safetySettings`.

**The new path** — add a provider through Settings → Providers → "Use a custom endpoint", pointed at an OpenAI-compatible service (`https://openrouter.ai/api/v1/chat/completions` with a model id such as `openai/gpt-4o`). Test connection should pass, and a story turn should generate. Expect the disclosure block to say the content rating is guidance rather than an enforced filter.

**Refusals worth confirming** — a `http://` endpoint or a local address is rejected at the verify step with a message explaining why, and never reaches `fetch`.

**Not run here:** no live multi-turn play session against a real non-Gemini provider, and no check of whether other models refuse prose Gemini writes today. The issue's own research comment argues that question wants a real test rather than a warning string; this PR ships the mechanism, not that evidence.

## Checklist

Docs weren't touched — the provider docs describe the config UI, which is unchanged in shape. Worth a follow-up once #895 turns the presets on.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CGJjxtUgXcgsBgAZ8Tve15)_


jerseycheese:
Pushed `6a03d56` addressing the CodeQL alerts and all three Codex findings. Flagging one judgment call that's yours to overrule.

**The SSRF finding was right, and worse than the guard assumed.** The check was string-only and ran in `resolveProvider`, far from the request. It blocked literal private addresses and nothing else — so a public hostname under an attacker's control whose A record points at `169.254.169.254` walked straight through, and `fetch` follows redirects, so even a genuinely public host could `302` to the same place with no re-check. These routes are unauthenticated and return the upstream body, so that's a full-response proxy, not a blind one.

Four changes:

- The guard now runs **at the sink**. `sendProviderRequest` takes `playerSuppliedEndpoint` and refuses before it can reach `fetch`, so no call path arrives having skipped it. Adapters declare whether their URL is player-supplied; Gemini's is a pinned constant and says `false`.
- **`redirect: 'error'`.** Following a redirect discards every check made before it.
- **Hostnames are resolved and every returned address checked** — the half a string check can't do. A name that won't resolve is refused rather than attempted.
- **Upstream error bodies** are no longer logged verbatim and are truncated before reaching the player (your P2 — a provider that echoes the request would put the bearer credential in that body).

**What this does not close, stated plainly:** resolve-then-connect leaves a DNS-rebinding window, because the address checked isn't the address the socket ends up using. Closing it needs the resolved IP pinned into the connection via a custom agent, which isn't available on every runtime this deploys to. That's in the `endpointGuard` header comment rather than left for someone to discover.

**The judgment call: I added an inline CodeQL suppression** at the `fetch`, with the reasoning inline. Bring-your-own-provider cannot exist without fetching a URL the player named, so CodeQL will flag this path no matter how well guarded it is. I'd rather that be an explicit annotation you can read and disagree with than an alert dismissed in the UI where the reasoning disappears. If you'd rather carry the alert open, or want the DNS-rebinding gap closed before this merges, say so and I'll change it — I didn't want to silently decide that on your behalf.

**Streaming refusals (P2) — good catch, real bug.** A `content_filter`/`error` final frame with no content was emitting a successful empty `done`, which `parseNarrativeResponse` then reported as a malformed response: an error about our own parsing for something that was the provider declining. It now emits an error event, matching what the non-streaming path already did. A refusal that produced partial prose still completes normally. Tested both ways.

Also fixed the polynomial-regex alert (unbounded `[A-Z]+` before an optional suffix, running on whole prompts — bounded to 5, since no rating is longer) and the Knip failure (`FinishReason` was declared and unused; the adapters and stream consumer now carry it, so the normalization contract is type-enforced rather than comment-enforced).

Full suite is 2776 passing across 405 suites; type-check, lint, stylelint, Knip, `deps:validate` and `deps:check` all clean with the dependency baseline unchanged at 17.

---
_Generated by [Claude Code](https://claude.ai/code)_

jerseycheese:
## One action needed from you: dismiss the SSRF alert

Everything else on this PR is green. The only remaining red is the code-scanning **alert** check — `CodeQL Analysis` itself passes; it's the alert gate that fails.

**Correction to my earlier comment:** I said I'd added an inline CodeQL suppression. That was wrong — GitHub code scanning doesn't honour inline `codeql[...]` comments (that's an LGTM feature), so it never suppressed anything. It's removed in `c9c6be6`, replaced with a comment that explains the acceptance and points here. I'd rather that than something that looks like a control and isn't.

### What to dismiss

[Alert #648 — Server-side request forgery](https://github.com/jerseycheese/Narraitor/security/code-scanning/648), `src/lib/ai/providers/core/request.ts`. Suggested reason: **"Used in tests"** doesn't fit; use **"Won't fix"** with a note along these lines:

> Accepted risk. The URL is player-supplied by design — this is bring-your-own-provider, and the feature cannot exist without fetching a URL the player named. Constrained by `assertPublicProviderEndpoint`: https only, private literals refused, hostname resolved and every returned address checked, redirects refused, response body bounded before it reaches the caller. Residual gap is DNS rebinding, documented in `endpointGuard`.

Dismiss it against the head commit once CI settles, so the fingerprint doesn't shift under you.

### Why this rather than a code fix

The alternative that would clear CodeQL is restricting outbound requests to an allowlist of known provider hosts. That genuinely eliminates arbitrary-host SSRF — but it contradicts #890's own acceptance criteria ("work with any provider that follows [the OpenAI-compatible standard]", "take the full URL from configuration") and would make the custom-endpoint form host-restricted. That's a product decision, so I didn't make it unilaterally. If you'd rather have the allowlist, say so and I'll switch — it's a contained change.

Worth knowing for that decision: these routes are unauthenticated apart from rate limiting, and they return the upstream response body, so this is a full-response proxy rather than a blind one. The controls above are what make that acceptable; they aren't a proof of unreachability.

---
_Generated by [Claude Code](https://claude.ai/code)_

jerseycheese:
@codex address that feedback -- please re-review, five commits landed since your last pass (f58a793d) covering the SSRF/CodeQL fixes plus a follow-up pass: non-Gemini providers were falling through to the server's Gemini key on five routes, the endpoint guard's IPv4-mapped IPv6 check could never match, and /api/ai/validate-provider had no rate limit.

Issue #1583, opened 2026-07-20
Retire the palette icon and the dead theme-menu CSS


jerseycheese:
## Plain Language Summary
The header still has a palette icon that opens a dropdown containing exactly one control, now that there's only one design system left. The chrome around it, and five CSS rules styling markup nothing renders any more, are all left over from the three-skin era.

## Current Feature
`src/components/Navigation/ThemeMenu.tsx`, a 70-line component rendering a `<Palette>` icon trigger with `aria-label="Appearance"` (`:57`). The open dropdown contains one child at `:63`: `<DarkModeToggle />`, which is a three-option radiogroup for Light, Dark, and System (`src/components/Navigation/DarkModeToggle.tsx:9-13`). All three options still do real work.

## Domain
- [x] Other: Design System / Navigation

## Enhancement Description
The behavioral half of the cleanup is already done. `src/lib/theme/index.ts` now exports only `ColorScheme`, with no `DesignSystem` type, no `THEMES` array, and no `theme` / `setTheme`. `ThemeSwitcher.tsx` was deleted in `03bedb40`, and #1546 shipped the rest through PR #1568.

What's left is the surface. A palette icon, a click-outside and Escape handler, and a 15rem dropdown sized for skin descriptions now wrap three icon buttons. The dead CSS is still in `workshop.css`, and none of these classes appear in any `.tsx` or `.ts`:

```css
/* :867 */ .theme-menu .header-dropdown-menu { width: 15rem; }   /* comment still says "let the skin descriptions wrap" */
/* :871 */ .theme-menu-item-text { min-width: 0; }
/* :875 */ .theme-menu-item-text > div { white-space: normal; }
/* :880 */ .theme-menu-mode { ... }                              /* comment still says "below the theme list" */
/* :886 */ .theme-menu-mode .dark-mode-toggle { ... }
```

It also produces an odd settings surface. `src/app/settings/page.tsx:71` renders the same collapsed `<ThemeMenu />` inside an "Appearance -> Color mode" card, so the control is hidden behind a palette icon inside a card that already labels what it does.

## Reason for Enhancement
A palette icon promises theme or color choices and delivers a light/dark toggle, which is a small "why is this here" moment on a surface every user touches. The dead CSS is worth deleting regardless of what happens to the icon, since it carries comments describing a feature that no longer exists and is exactly the kind of thing that gets copy-pasted before someone notices.

## Possible Implementation
Two options, and the second is probably right for settings at minimum.

[...]

Issue #1578, opened 2026-07-20
Epilogue prompt permits cliffhangers on every non-tragic ending


jerseycheese:
## Plain Language Summary
Endings can finish on a cliffhanger, which undercuts the one moment the whole session builds toward. The prompt is asking for that, so it's a prompt fix rather than a code fix.

## Current Feature
Ending and epilogue generation, `src/lib/promptTemplates/templates/endingTemplates.ts`, rendered through `src/lib/ai/endingGenerator.ts:166`. There's a single prompt path for all four tones.

## Domain
- [x] Narrative Engine

## Enhancement Description
The epilogue instruction is thin on resolution. Line 67 asks the model to "provide narrative closure while respecting the ending type", and then the global closing checklist at line 108 undercuts it with "provide closure while leaving room for imagination".

The tone menu goes further and asks for open-endedness outright, lines 52-53:

> MYSTERIOUS: Use when there are unresolved questions, ambiguous outcomes, enigmatic events, or the story leaves mysteries unsolved. The ending raises new questions.

Every anti-cliffhanger guardrail in the file is scoped to death. Lines 111-118 forbid "what lies ahead" and "leave open the possibility of continuation", but that block is prefixed `DO NOT (especially for tragic/fatal endings):` and closes with `For tragic endings, be clear: the story has ended.` `prepareEndingTemplateVariables` (`:149-167`) does the same, injecting `fatalContext` only when `desiredTone === 'tragic'`.

So nothing in the template requires plot resolution for the three non-tragic tones, and the playtest drew `mysterious`, which asked for exactly the cliffhanger it got. Worth noting the tone was genuinely model-chosen, since both fallback paths hardcode `'hopeful'` (`endingGenerator.ts:79` and `:106`).

## Reason for Enhancement
The ending is the payoff for a whole session, and a quarter of the tone space currently instructs the model to withhold it. Cheap to fix, and it lands on the moment players are most likely to remember.

## Possible Implementation
Add a closure requirement to the EPILOGUE block that applies to every tone, then reframe MYSTERIOUS so the ambiguity is thematic rather than plot-level, something along the lines of resolving the immediate arc while leaving lingering questions about meaning rather than about what happens next. Line 108's "leaving room for imagination" wants rewording or dropping.

Prompt text only, no code or type changes.

## Alternatives Considered
[...]

PR #1787, opened 2026-08-14
Assert layout invariants instead of leaning on baselines alone

jerseycheese:
## Description

The visual suite is 42 `toHaveScreenshot` calls plus some presence checks, so a layout rule can stop applying without any framed pixel changing. Three of the v1.0 playtest regressions got through exactly that way. This lands the three pieces of #1587 that weren't already covered by the per-issue assertions in `manuscript-regression-assertions.spec.ts`.

**The two-tier convention** goes into `public_docs/development/visual-testing-best-practices.md` as a new "Two Tiers of Layout Assertion" section. That file is the one that already answers "what should this test actually check" (it holds the anti-patterns and the split-threshold strategy), where `visual-regression-testing.md` is the how-to-run guide and `visual-test-examples.md` is a snippet library. The section covers tier 1 (geometry and behavior, the default), tier 2 (resolved-value tripwires, only where the cascade decides the winner and never on their own), why a computed-style assertion mostly doesn't work, and what to do when a fix has no runtime invariant at all.

That last part matters more than it looks. #1578 is prose quality and #1583 is mostly deletion, so neither has anything to measure. Without an answer for that case, the first awkward fix turns into a fake assertion that reads as coverage.

**The eslint guard** is `no-restricted-syntax` with two AST selectors, added to the existing test-file block in `eslint.config.mjs` alongside the `jest/*` rules from the rigged-test cleanup. It flags `scrollHeight`, `clientHeight`, `scrollTop`, `offsetHeight`, `offsetTop` and `getBoundingClientRect` being defined on an element in a test file, in both the `Object.defineProperty` and the direct-assignment form. No custom plugin, no new package.

It's `error` where its neighbours are `warn`, because those landed against an existing backlog and needed phasing in. This one starts at zero, since the third piece clears it.

**Three jsdom files move to Playwright.** Stubbing layout geometry means the test isn't observing layout, and jsdom leaves no other option, so these get moved rather than repaired.

| File | Moved | Stayed |
|---|---|---|
| `NarrativeHistory.streaming.test.tsx` | 2 ResizeObserver anchoring tests | 8 tests: buffered rendering, observer setup/teardown, streaming preview states, layout classes |
| `NarrativeHistory.readingPosition.test.tsx` | all 4 (file deleted) | none |
| `useTutorialTooltipReposition.test.tsx` | 1 reflow-moved-the-target test | 3 tests: listener wiring, beacon-wrapper filtering, teardown when the tour stops |

Two of the four `readingPosition` invariants (a scrolled-up reader isn't moved, and "jump to latest" takes them down) were already covered by `manuscript-regression-assertions.spec.ts`, so they aren't duplicated. The other two, plus both anchoring cases, are new in `tests/visual/narrative-reading-position.spec.ts`.

One assertion is dropped rather than moved: the "a plain scroll does not report the target as moved" half of the tutorial test. In jsdom `window.scrollY` never changes, so that half passed for the wrong reason, and there's no user-visible thing to measure in a browser.

## Related Issue

Refs #1587. The base here is `develop`, so a `Closes` keyword wouldn't fire - this needs closing by hand after merge.

## Type of Change
- [x] Documentation update

## TDD Compliance
- [x] All new code is tested
- [x] All tests pass locally

Not TDD in shape: there's no product code here, only tests, a lint rule and a doc. The new specs were validated the other way round, against a deliberately broken build (see Implementation Notes).

## User Stories Addressed

N/A. Internal testing convention, no player-facing behavior.

## Flow Diagrams

N/A.

## Component Development

N/A. No components changed.

## Implementation Notes

Both new specs were checked against a deliberately broken build before being trusted, since a green assertion proves nothing about what it would catch.

- Commenting out `popperRef.current?.update()` and the `onTargetMoved` call in `useTutorialTooltipReposition` puts the tooltip 160px off its target: `Expected: <= 8, Received: 160`.
- Disabling the follow-new-segment branch, the deferred initial scroll, and the ResizeObserver anchoring in `NarrativeHistory` turns all four reading-position tests red.

Everything was restored afterwards; the diff contains no product source changes.

The reading-position spec measures `.manuscript-overlay-main`, which is the element the app actually scrolls. The jsdom tests were driving the Radix viewport, which `NarrativeHistory` only falls back to when the play surface isn't in the tree - so they were stubbing geometry onto an element the app never scrolls, which is why they survived the DS3 auto-scroll regression intact.

The content-growth cases use a viewport narrowing to rewrap the prose. That's real reflow rather than a fabricated resize, and it's the same shape as a segment growing while it streams in. `updateSegment` was tried first and doesn't re-render the rendered segment, so it isn't a usable driver.

Neither new spec captures a screenshot, so nothing here needs a baseline.

## Screenshots

N/A. No visual changes.

## Visual Baseline Changes

None. No files under `tests/visual/**-snapshots/**` are touched.

N/A.

N/A.

## Quality Checks
- [x] Linting passed
- [x] Type checking passed
- [ ] Security audit passed
- [x] No unhandled promises

Security audit wasn't run - nothing here touches dependencies or runtime code.

## Testing Instructions

The four gate commands, all exit 0 locally:

```
npm test           # 396 suites, 2677 tests passed
npm run type-check
npm run lint       # 0 errors, 149 pre-existing warnings
npm run lint:css
```

The new Playwright specs, against a dev server on this worktree's port:

```
npx playwright test tests/visual/narrative-reading-position.spec.ts --project=chromium
npx playwright test tests/visual/tutorials/tooltip-reposition.spec.ts --project=tutorials
```

Both pass, and both were re-run repeated to check for flake (4x and 3x respectively).

To see the eslint guard fire, add `Object.defineProperty(el, 'scrollHeight', { value: 1000 })` to any test file and run `npm run lint`.

## Checklist

Accessibility: N/A, no UI changes. One new warning does appear in `npm run lint` - a `no-explicit-any` on the store handle in the new spec, matching how every other spec in `tests/visual/` reaches for `window.useNarrativeStore`.


jerseycheese:
Two follow-ups since the first push, both worth recording.

**The geometry guard had a hole.** The `Object.defineProperty` selector covered six properties, the direct-assignment selector covered only `getBoundingClientRect`. So `(element as any).scrollHeight = 1000` still landed clean, costing nothing but a non-blocking `no-explicit-any` warning, which is the exact stub the streaming test was using before it moved. Both selectors now cover the same six. Verified against a throwaway file: all six error in that cast form, and `title` and `dataset` stay quiet.

**One of the new assertions was vacuous, and a mutation probe caught it.** Dropping the anchoring's own guard in `NarrativeHistory` so the ResizeObserver scrolls unconditionally does yank a scrolled-up reader to the bottom (scrollTop 1123), and "Leaves a reader who scrolled up where they are" stayed green through it. scrollHeight grows during layout but the anchoring runs afterwards off the observer, so reading the position the moment the page got taller lands in the frame before the wrong scroll arrives.

It now waits for the scroller's geometry to hold still across ten consecutive animation frames, and fails with 1123 against an expected 0. `waitForFunction` polls on rAF, so that's frames rather than a wall-clock sleep. The probe itself was reverted; there are no product-source changes in this PR.

Mapping for the two anchoring assertions that came out of `NarrativeHistory.streaming.test.tsx`, since it's not obvious from the file names:

| jsdom test removed | Playwright replacement | Mutation evidence |
|---|---|---|
| `anchors on ResizeObserver updates when user has not manually scrolled` | `Keeps the latest beat in view when a reflow makes the story taller` | Disabling the anchoring alone: 543px from the bottom, expected <= 4 |
| `does not force-scroll on ResizeObserver updates when user has scrolled away from bottom` | `Leaves a reader who scrolled up where they are when a reflow makes the story taller` | Removing the guard so it anchors unconditionally: scrollTop 1123, expected 0 |

jerseycheese:
Follow-up on the wall-clock waits in the tooltip spec. Two of the three were doing what `waitForReflowToSettle` already handles properly in the narrative spec, so they're gone.

`readSettledAnchorOffsets` now waits for the whole anchor signature - the target's document position plus both offsets - to hold still across ten consecutive animation frames. `waitForFunction` polls on rAF, so a loaded runner stretches the real time that covers rather than running out of it, which is the failure mode a fixed 500ms has here. One rule now covers the initial Joyride placement and the post-reflow redraw, where before each had its own sleep.

Checked both directions afterwards, since a wait that returns early would quietly turn this into exactly the vacuous assertion this PR is about:

| | Result |
|---|---|
| Hook intact, 4 repeats | green |
| `popper.update()` and `onTargetMoved` commented out, 2 repeats | red both times, 160 against an expected 8 |

The remaining 600ms waits out the wizard's own auto-start timer rather than a layout signal, and matches what the tour screenshot specs already do, so it stays.

Gates after the change: `npm test` 0 (396 suites / 2677 tests), `npm run type-check` 0, `npm run lint` 0, and the Tutorial Visual Tests job came in at 2m52s against 3m38s before.

Issue #1525, opened 2026-07-10
Build a trigger-eval harness to score the skill library's routing accuracy


jerseycheese:
## Plain Language Summary

Each of the 16 skills in `.claude/skills/` ships a `trigger_eval.json` file — a set of ~23 example queries labeled with which skill should (or shouldn't) fire. Nothing actually runs them. We should build a small harness that feeds those queries through real skill-trigger routing and scores how often the right skill is picked, so we can catch trigger regressions and false positives.

## Current State

Every `narraitor-*` skill has `evals/trigger_eval.json` (16 files, ~370 labeled queries total). Schema per entry:

```json
{"query": "where do the zustand stores live in this repo?", "should_trigger": true, "expected_skill": "narraitor-repo-orientation"}
```

They were authored to encode each skill's trigger boundaries (positives, sibling near-misses that should route elsewhere, and null cases that should fire nothing) — but they've never been executed against an actual trigger mechanism. So we have the ground-truth labels and no measurement. This is called out as an open gap in `.claude/skills/_uncertainty_register.md` ("the trigger evals have not been run through an actual trigger harness") and `_maintenance_plan.md` (re-run after description edits).

The transfer benchmark that shipped with #1513 measured *usage* quality once a skill is loaded (treatment 27/28 vs baseline 21/28). It did NOT measure *triggering* — whether the right skill activates from a cold query. This issue covers that second axis.

## Domain

- [x] Other: agent tooling / `.claude/skills/`

## Enhancement Description

Build (or wire up) a harness that:
1. Loads all `.claude/skills/narraitor-*/evals/trigger_eval.json` files.
2. For each query, routes it through the real skill-trigger path and records which skill(s) fire.
3. Scores against the labels: precision/recall per skill, plus a confusion matrix for the sibling near-miss cases (the most valuable signal — `_trigger_matrix.md` lists the confusable pairs to watch).
4. Emits a report (pass rate, false-positive and false-negative lists) so a description edit that breaks routing is visible.

Runnable ad hoc, and cheap enough to re-run whenever a skill description changes.

## Reason for Enhancement

Trigger accuracy is the whole point of the frontmatter descriptions — if the wrong skill fires (or none does), the encoded doctrine never reaches the session. Right now we have 370 labeled cases sitting unused. A harness turns them from dead fixtures into a regression gate, and the `_maintenance_plan.md` already assumes one exists ("re-run the confusable pairs after any description edit").

## Possible Implementation

[...]

PR #1788, opened 2026-08-14
Score the skill library's trigger routing instead of guessing at it

jerseycheese:
## Description

The skill library ships 378 labelled trigger queries across 16 `narraitor-*` skills and nothing had ever executed them, so `_uncertainty_register.md` carried it as an open gap while `_maintenance_plan.md` already told you to "re-run the confusable pairs after any description edit" against a harness that did not exist. This adds the harness.

`npm run skills:trigger-eval` loads every `.claude/skills/narraitor-*/evals/trigger_eval.json`, sends each query to a real non-interactive Claude Code session in this repo, records which skill the model reaches for, and scores it: overall pass rate, per-skill precision and recall, a confusion matrix over the sibling near-misses, and explicit false-positive and false-negative lists with the offending queries printed in full.

The part worth reading carefully is how routing gets invoked, because faking it would have been easy and the fake numbers would have looked fine. There is no callable router in Claude Code to import. Triggering is a model decision made over the frontmatter `description` fields, so the harness drives that decision for real:

```
claude -p "<query>" --output-format stream-json --verbose --max-turns 1 \
  --no-session-persistence --setting-sources project \
  --exclude-dynamic-system-prompt-sections [--model M] [--tools Skill]
```

Run from the repo root with `CLAUDECODE` stripped from the environment. The first tool call of the first assistant message is the observation: `Skill` means that skill fired, a `Read` of `.claude/skills/<name>/SKILL.md` counts too, prose or any other tool means nothing fired. The candidate set is the repo's own skills, the descriptions are the real ones on disk, the decision is the real model's. Nothing in the harness reads a description or matches a keyword.

`--setting-sources project` is doing two jobs. It pins the candidate set to the skills these fixtures label so the score reproduces on another machine, and it stops every single query from firing the global `SessionStart` hooks, one of which reaps stale worktrees. Firing that a few hundred times over would have been destructive.

### The two observation modes, and why neither is ground truth

This is the honest limitation and it is surfaced in the script header, in the register, and in the footer of every report the tool prints.

| Mode | What the session gets | Which way it is wrong |
|---|---|---|
| `forced-choice` (default) | The `Skill` tool and nothing else | Reads high. The model cannot answer by going and looking, so picking a skill is its only move besides prose. |
| `session` | The normal toolset | Reads low. The model routinely opens with a context-gathering `Bash` call and only reaches for the skill on a later turn, which the single-turn window scores as silence. |

Both were confirmed by hand before being written up. In session mode `"which test tiers does this loreStore change actually need to run"` fires `narraitor-validation-and-qa` on the second tool call, after an opening `git status`, so the first-action rule scores a real hit as a miss. In forced-choice mode `"write me a python script to rename files in a folder"` fires `narraitor-repo-orientation`, which it would not have done with somewhere else to go. Truth sits between the two, and the report says so rather than picking the flattering number.

### Actual numbers

Yes, this has been run end to end against real queries from a local macOS checkout, and the numbers below are observed, not estimated. Worth stating plainly because a fair question came up: until the errored-case work landed, a crashed run and a clean run looked identical, so a "successful" run could in principle have been nothing but silent failures. It was not. The verified run records `errored: 0`, `failedSessions: 0`, 96 of 96 cases scored, and 21 distinct observed labels including skills outside this library (`dead-code-cleanup`, `run`, `claude-api`). A router that never answered could not produce any of that. The known failure mode where a `claude -p` child fails when spawned inside a Claude Code session did not apply here.

96 queries, six per skill, stratified across positives, near-misses and nulls. Model `sonnet`, one run per query.

| | forced-choice (verified) | forced-choice (earlier) | session |
|---|---|---|---|
| Overall | 82/96 (85.4%) | 86/96 (89.6%) | 71/96 (74.0%) |
| Positives | 24/32 (75.0%) | 28/32 (87.5%) | 20/32 (62.5%) |
| Near-misses | 29/32 (90.6%) | 30/32 (93.8%) | 19/32 (59.4%) |
| Null cases | 29/32 (90.6%) | 28/32 (87.5%) | 32/32 (100%) |

The first two columns are the same 96 queries, same mode, same model, run twice. They differ by 4.2 points. That is the nondeterminism the report warns about, now measured rather than asserted, and it sets the noise floor: under about five points, do not read a delta as a regression. The register and the maintenance plan both carry that number now.

The session-mode column was produced before the `--tools-mode` flag existed, on code that passed no `--tools` argument, which is what `--tools-mode session` does now. The routing invocation was identical, only the metadata label differs.

In session mode there were zero false positives. Every failure was silence, never a sibling stealing the query. Whatever else is true of these descriptions, they are not colliding with each other.

The verified forced-choice run found ten misroutes, and they are the useful output of the whole exercise:

```
2  narraitor-hardest-problem-campaign -> none
1  narraitor-build-test-env           -> dead-code-cleanup
1  narraitor-change-control           -> narraitor-prompt-template-governance
1  narraitor-change-control           -> none
1  narraitor-docs-and-writing         -> none
1  narraitor-domain-reference         -> narraitor-debugging-playbook
1  narraitor-hardest-problem-campaign -> narraitor-change-control
1  narraitor-product-frontier         -> none
1  narraitor-repo-orientation         -> none
1  narraitor-storybook-app-parity     -> narraitor-debugging-playbook
```

Two of those reproduce across both forced-choice runs and are the ones I would chase first. `"where do the zustand stores live in this repo?"` is the flagship positive from the library's own README example and it does not fire. `storybook-app-parity -> debugging-playbook` on `"works in storybook but blank in the app"` is one of the exact confusable pairs `_trigger_matrix.md` flags as highest-risk, now measured rather than assumed.

Three of the four null-case false positives went to built-in Claude Code skills (`run`, `artifact-design`, `claude-api`), not to anything in the narraitor library. Worth knowing before reading the null-case number as a library problem.

Per the brief this is read-only. No skill description was touched. Tuning off these results is a separate follow-up.

### Did skill-creator's eval tooling fit

Checked properly, and the answer is a partial yes that does not go far enough.

`skill-creator/scripts/run_eval.py` is genuinely a trigger-eval runner, and it consumes `[{query, should_trigger}]` and ignores extra keys, so it would read these fixtures without modification. What it cannot do is the thing the issue calls the most valuable signal. It evaluates one skill in isolation: it writes a synthetic command file containing only that skill's description into `.claude/commands/`, runs `claude -p`, and returns a boolean for whether that one skill fired. The other nineteen skills are never in the candidate set, so a near-miss labelled "should route to `narraitor-debugging-playbook`" can only ever be scored "did not fire", never "fired the right sibling". No confusion matrix is reachable from that shape, and cross-skill collisions cannot occur in a candidate set of one. It also tests a fabricated description-only stub rather than the real skill.

`references/schemas.md` `evals.json` is a different thing entirely: prompt plus expectations, graded by a subagent for usage quality once a skill is already loaded. That is the axis the transfer benchmark already covered.

So the local scorer got written, but the approach was lifted rather than reinvented. Stripping `CLAUDECODE` to nest `claude -p`, watching the stream for the `Skill` tool, and voting across repeated runs all come from `run_eval.py`.

### Fixture observations, not fixed here

Out of scope per the brief, noted for a follow-up:

- The issue says roughly 370 queries. Actual count is 378 entries, 376 unique. Two queries appear in two files each, with consistent labels, and the harness collapses them.
- `narraitor-architecture` and `narraitor-pattern-alignment-skill` have no `evals/` directory, so 16 of the 18 `narraitor-*` skill folders carry fixtures. Both still appear as `expected_skill` in five entries elsewhere and get scored without owning a file.
- The null cases assume nothing fires, but built-in Claude Code skills are in the real candidate set and legitimately claim some of them. A query like "write a python script to rename files in a directory" may never be satisfiable as "none" in a real session. Whether those cases should be relabelled is a call for the owner.

## Related Issue

Refs #1525. The base here is `develop`, not the default branch, so a `Closes` keyword would not fire. Needs closing by hand after merge.

## Type of Change
- [x] Documentation update

## TDD Compliance
- [ ] All new code is tested
- [x] All tests pass locally

Two boxes left honestly unchecked. The scoring maths was written first and the test second, not the other way round. And the coverage is deliberately partial: `trigger-eval-lib.cjs` is fully unit-tested because it is pure, while the CLI's process spawning and stream parsing are not, since testing them means either mocking `claude` into meaninglessness or spending money in CI. That split is why the maths lives in its own file, and why the errored-case handling was moved into the scorer rather than left in the CLI. Fifteen tests, three of them on the errored path, including the one that matters: a case the model was never asked about must not count against a skill's recall.

## User Stories Addressed

N/A. Agent tooling under `.claude/skills/`, no product surface.

## Flow Diagrams

N/A.

## Component Development

N/A, no UI.

## Implementation Notes

The split follows the existing `ds-canon-lib.cjs` / `verify-ds-canon.cjs` pattern: pure logic in a library with no fs, no spawn and no process, a CLI wrapper that does the I/O, and the unit test against the library. Scoring works on plain data, so the maths can be checked against a six-case fixture in under a second and for nothing.

A misroute is deliberately counted in both the false-positive and the false-negative list, because it is both: the expected skill stayed silent, and a sibling fired that should not have. That is what makes the per-skill precision and recall arithmetic add up, and the test pins the behaviour. Cases with no observation recorded come back as `unresolved` rather than being scored as misses, so an aborted run cannot masquerade as a routing failure.

A case has three outcomes, not two: a skill fired, nothing fired, or the router was never reached. That third one is deliberate and it is the reason the scorer owns it rather than the CLI. Collapsing "could not ask the model" into "the model chose wrong" is exactly how a harness ends up printing confident numbers with nothing behind them, which is the thing this script's own header warns against. Errored cases are excluded from precision, recall, the pass rate and the confusion matrix, listed at the bottom with their cause, announced in a banner at the top, and carried in the exit code: 0 when everything scored, 2 when some cases errored, 1 when nothing scored at all.

Cost drove several decisions. Every query is a live session, so a full 376-query sweep is real money and a good half hour. `--sample N` takes a stratified slice, `--skill` and `--pairs` scope to what actually changed, `--dry-run` prints the plan without spending anything, and `--out` plus `--from` lets you rescore saved observations for free. The 96-query forced-choice run above reported a 4.63 USD floor, billed by only 30 of 96 sessions, since a session that reaches a skill gets killed at the decision and never emits a result event. Real spend was higher than the floor and the report labels it as a floor rather than a total.

`--runs N` votes across repeats for the nondeterminism. Default is 1 to keep the cheap path cheap.

`_trigger_results/` is gitignored. A committed results snapshot is stale the moment a description changes, and the point of this tool is that you re-run it rather than read an old file.

Docs touched: `_uncertainty_register.md` narrows the open gap to what is still genuinely unmeasured (mid-conversation triggering, other models) instead of closing it outright, `_maintenance_plan.md` names the command in its drift check and its "updating trigger evals" section, and `_trigger_matrix.md` gets one line pointing at `CONFUSABLE_PAIR_SKILLS` so the `--pairs` list and the documented pairs do not drift apart.

## Screenshots

N/A.

## Visual Baseline Changes

N/A.

N/A.

## Quality Checks
- [x] Linting passed
- [x] Type checking passed
- [ ] Security audit passed
- [x] No unhandled promises

Security audit not run, no dependency change in this PR. The `console.log` calls in the new script are the report itself, in `scripts/`, where the ESLint config turns `no-console` off for exactly that reason.

## Testing Instructions

Costs nothing:

```bash
npm test -- scripts/__tests__/trigger-eval-lib.test.js
npm run skills:trigger-eval -- --dry-run
npm run skills:trigger-eval -- --dry-run --sample 3
```

The dry run should report 16 fixture files and 376 unique queries, and confirm the fixtures parse.

Spends tokens, so start small:

```bash
npm run skills:trigger-eval -- --skill narraitor-debugging-playbook --sample 3 --model sonnet
npm run skills:trigger-eval -- --pairs --model sonnet --out /tmp/pairs.json
npm run skills:trigger-eval -- --from /tmp/pairs.json          # rescore, free
npm run skills:trigger-eval -- --sample 6 --model sonnet --tools-mode session
```

Full gate, each run separately, all exit 0:

```bash
npm test           # 398 suites, 2696 tests
npm run type-check
npm run lint       # 0 errors, 148 pre-existing warnings
npm run lint:css
```

## Checklist

`run-trigger-eval.cjs` is 424 lines, over the 300 guideline. Roughly 60 of those are the header block spelling out what "real routing" means and where each observation mode is wrong, which seemed worth more than the line count. It sits alongside `validate-prompt-templates.js` at 368 and `github-label-creator.js` at 460, so it is not out of step with the rest of `scripts/`. Happy to split the router out if you would rather hold the line.

Accessibility: N/A, terminal output only, and it is plain ASCII with no symbols standing in for meaning.


jerseycheese:
All three Codex findings were legitimate and are fixed in 9f7e6993.

The P1 was the important one, and it is the exact failure mode this harness is supposed to be immune to. If `claude` was missing, unauthenticated, or rejected an option, the child exited quietly with `observed` still set to `none`, which is indistinguishable from a genuine "no skill fired". A completely broken environment would have printed a confident 0% and read as a description problem. A session now has to produce a real model turn to count as an observation. Failed ones are collected separately, excluded from scoring so they land in the already-tested `unresolved` bucket rather than counting against a skill that never got a fair shot, and the count prints in the report header. If nothing answers at all the run exits 1.

Verified by running it with `claude` off the PATH:

```
2/2 sessions failed to produce a turn:
  could not run claude: spawn claude ENOENT - "where do the zustand stores live in this repo?"
Those queries are excluded from scoring, not counted as misses.

No query produced a model turn. Check that the claude CLI runs here before trusting any score.
exit 1
```

The `--from` finding was right too, and it broke the exact flow the testing instructions document. The saved case list now rides along in the output file and rescoring uses it. A `--skill narraitor-storybook-app-parity --sample 3` run scoring 2/3 now rescores as 2/3 through a bare `--from`, where before it would have scored 2 against all 376 cases and called the other 373 unresolved.

Numeric validation is in as well. `--sample abc` and `--runs 0` both exit 2 before anything spawns, which matters because sampling is the only thing standing between a typo and a full paid sweep.

Gate after the fixes, each run separately: `npm test` 0 (398 suites, 2696 tests), `npm run type-check` 0, `npm run lint` 0 (148 pre-existing warnings, no errors), `npm run lint:css` 0.

One thing I did not do: add unit tests for these three. They all live in the CLI wrapper, in process spawning and argument parsing, and testing them properly means either mocking `claude` into something that proves nothing or spending money in CI. Each was verified by execution instead, with the output above. The scoring maths that the P1 fix routes into, the `unresolved` path, was already covered.

jerseycheese:
Went back at the P1 harder, because the first fix was not enough.

Excluding failed sessions from scoring stopped them counting as misses, but a case still only had two possible outcomes: a skill fired, or nothing fired. "The router was never reached" got filed under the second one. That is the exact failure this script's own header warns about, so it now has a third outcome of its own, tracked in the scorer rather than the CLI so it is testable.

Errored cases are excluded from precision, recall, the pass rate and the confusion matrix. They print in a banner at the top, get their own section at the bottom with the underlying cause, and drive the exit code: 0 when everything scored, 2 when some cases errored, 1 when nothing scored at all. A fully broken environment now looks like this instead of a clean zero-recall table:

```
*** WARNING: the router could not be reached for some queries. ***
*** 2 of 2 cases errored. They are excluded from every
*** number below, and are NOT counted as routing misses.

NO CASE WAS SCORED. Nothing below is a measurement of routing quality.
exit 1
```

Three new tests cover it, including the one that matters: a case the model was never asked about must not count against a skill's recall.

**On whether this has actually run.** Fair question given the bug, and the answer is yes. Re-ran the full 96-query sample under the fixed code and the artifact records `errored: 0`, `failedSessions: 0`, 96 of 96 scored, 21 distinct observed labels including skills outside this library (`dead-code-cleanup`, `run`, `claude-api`). A router that never answered could not produce that spread. The known failure mode where a `claude -p` child dies when spawned inside a Claude Code session did not apply here.

The verified run scored 82/96 (85.4%), against 86/96 (89.6%) for the identical sample in the identical mode earlier. Same queries, same model, 4.2 points apart. That is the nondeterminism the footer warns about, now measured, and it sets the noise floor: under about five points, do not read a delta as a regression. Both the register and the maintenance plan carry that number now. The PR body has the full table and the ten misroutes.

Gate, each run separately: `npm test` 0 (398 suites, 2699 tests), `npm run type-check` 0, `npm run lint` 0, `npm run lint:css` 0.

Issue #1206, opened 2026-05-10
[BUG] markSessionEnded throws ReferenceError: require is not defined on story end


jerseycheese:
## Plain Language Summary
When a player ends a story (triggering the EndingScreen), `narrativeStore.markSessionEnded` throws `ReferenceError: require is not defined`. The UI still renders correctly, but session lifecycle status may not propagate to the saved session record.

## Description
Discovered during a post-#1081 flow audit. The error fires silently — the EndingScreen renders all sections (epilogue, achievements, legacy, etc.) without visible breakage — but the underlying state mutation likely fails. Potential data integrity issue if downstream code expects the session lifecycle status to be flipped.

## Steps to Reproduce
1. Start any play session
2. Make at least one choice to advance the narrative
3. Click "End Story"
4. Wait for EndingScreen to render
5. Open browser devtools console — see error from `narrativeStore.ts` `markSessionEnded`

## Expected Behavior
- `markSessionEnded` completes without errors
- Session lifecycle status is updated and persisted

## Actual Behavior
- Console shows: `Failed to propagate session lifecycle status on ending ReferenceError: require is not defined`
- Source: `src/state/narrativeStore.ts` (`markSessionEnded`)
- EndingScreen still renders, but state mutation silently fails

## Environment
- Browser: Chromium via bdg
- OS: macOS
- Theme: DS1 (theme-agnostic — code path is in store, not UI)
- Branch state: post-#1081 merge (develop)

## Additional Context
- Almost certainly a CommonJS `require()` accidentally left in code that runs in the browser/ESM context
- Quick fix likely: replace the offending `require()` with an `import` (or hoist the import to the module top)
- Likely pre-existing, not a #1081 regression — this is a code-path bug, not a styling one
- Suggested investigation: `grep -n "require(" src/state/narrativeStore.ts` and the dynamic import path that fires from `markSessionEnded`

jerseycheese:
Done. Merged in PR #1209 (958229ec). The eval('require()') calls are gone, replaced with static ESM imports — no more silent ReferenceError on markSessionEnded. Regression test added; full 2139-test suite passes.

PR #1810, opened 2026-08-15
Add comment and test audit scripts, reap worldStore's restating comments

jerseycheese:
## Description

Adds `npm run audit:comments` and `npm run audit:tests`, the detection half of the new `comment-reaper` and `test-reaper` skills, plus the first reap as a dogfood.

The reason these exist rather than an extension to `dead-code-cleanup` or `doc-rot`: those skills key off markers this repo has already cleared. Run their criteria against `src/` and you get zero snapshot tests, zero `.skip`, zero orphaned test files, zero commented-out code, and one untracked TODO. The sweep finds nothing and stops.

What is actually here is a different kind of debt. Comments that are accurate and still worthless, and tests that pass no matter what the code does:

| Finding | Count |
|---|---|
| Issue/PR-number citations in comments | 214 (139 prod, 75 test) |
| Comments restating the line below them | 198 |
| Redundant one-line docblocks | 112 |
| Archaeological comments | 16 |
| Section-divider banners | 14 |
| Tests whose every assertion is a mock call-check | 274 (31 with a name promising more) |
| Files that are mostly single-assertion render checks | 44 |
| Cases asserting only `toBeDefined`/`toBeTruthy` | 2 |
| Duplicate test names within one describe block | 0 |

Both scripts follow `audit-css.mjs`: no new dependencies, write nothing, delete nothing, always exit 0. Output is grouped into the tiers the skills act on, so tier 1 can be applied mechanically and the rest gets read first.

Neither script is a parser, deliberately. The restating check compares a comment's words against the identifiers on the next line. The mock-only check looks at assertion shape rather than coverage, because "was this line executed" is the wrong question: a mock-only test executes plenty and asserts nothing about any of it. Both are tuned to under-report. A missed noise comment costs nothing; a false positive costs trust in the whole list.

## Related Issue

None. This came out of a request to build the two routines rather than from a filed issue.

## Type of Change

- [ ] Documentation update

## TDD Compliance

- [ ] All new code is tested
- [x] All tests pass locally

Not TDD, and marking that honestly. These are advisory audit scripts whose correctness question is "does this finding hold up when you read the source", which is answered by sampling real output, not by unit tests. That sampling is in the Testing Instructions below and it did find three defects. The `worldStore` change is comment deletion with no behavior to test.

## User Stories Addressed

None directly. This serves the comment hygiene rules in `CLAUDE.md`, which flag issue-number citations as the most-violated item and had nothing looking for them.

## Flow Diagrams

N/A.

## Component Development

N/A, no UI in this PR.

## Implementation Notes

Findings sort into three tiers, and tier 3 (never touch) is the load-bearing one:

- A regression test's `#1206` is the reason the test looks strange. Strip the number and the next person deletes the test. Citations inside test files are reported in their own section for exactly this reason.
- The field-level docblocks in `src/types/*.types.ts` are the closest thing the repo has to a data-model spec, and IDE hover reads them. `typeFieldJsdoc` buckets separately from `redundantJsdoc` so a sweep cannot mix them up.
- Mock-only tests are flagged "rewrite or delete, you pick" and never auto-deleted. They mark a coverage gap, and deleting one hides the gap instead of closing it.

One repo-specific trap the test skill calls out: `jest.setup.ts` applies a global `jest.mock('@/state/worldStore')`. Any store test runs against an auto-mock suite-wide, so a mock assertion is sometimes the only observation available. Do not judge one without accounting for that.

Calibration was done by sampling, and it moved real numbers. Three shapes were misfiring, all the same mistake of reading a present-tense phrase as history:

- Group headers. `// Sci-fi` over four `scifi-*` preset entries labels the run rather than restating one line. A comment that also matches the line after next is now treated as a group header. Restating went 229 to 198.
- Adjectival "previously". In "any previously encrypted data", previously describes the data, not the code's history. An article in front is the tell.
- Bare "extracted from". "JSON is extracted from the response" describes what the code does. Now requires was/were/been in front, which still catches "Persistence was extracted from the provider".

Redundant JSDoc stayed at 112. Fourteen spread across the list were checked and every one was genuine.

Codex review found four blind spots in `audit-tests.mjs`, all verified against the source before fixing and all real (`6e0dad9b`). Two were coverage holes: the audit only scanned `src`, missing the seven suites under `__tests__/` and `scripts/` that `jest.config.cjs` also matches, and multiline `it.each([` declarations were invisible because the case name sits on the `])(` line. Two were detection defects: a bare `assert` token did not match named helpers like `assertChoicesVisible()`, and mock-only detection required the subject to *look* mock-named, so `expect(notFound).toHaveBeenCalled()` slipped through purely on naming. A jest call matcher only works on a mock or spy, so the matcher already proves the subject.

Net effect: case count 2482 to 2580, mock-only 173 to 274, and `noAssertion` 3 to 0 because all three were false positives. Duplicate names are 0 and always were; the earlier "3" in this description was the tier-1 total, which those three false positives made up. Duplicates are scoped per describe block on purpose, since jest reports the full `describe > it` path and the same `it` text under two different describes is unique.

The two scripts are 370 and 318 lines, over the checklist's 300-line limit, so that box is unchecked. `scripts/` already holds four larger files (`github.js` at 572, `generate-homepage-showcase.mjs` at 561), so the limit does not appear to apply to this directory in practice. Splitting either script would separate the pattern definitions from the single pass that uses them for no real gain.

Companion skills live in the marketplace repo (`jerseycheese/agent-skills`, commit `e36a0ed`), along with two report-only weekly sweeps under `~/.claude/scheduled-tasks/`. Nothing in this PR depends on them; the scripts stand alone.

## Screenshots

N/A.

## Visual Baseline Changes

N/A.

N/A.

## Quality Checks

- [x] Linting passed
- [x] Type checking passed
- [ ] Security audit passed
- [x] No unhandled promises

Security audit not run; no dependencies were added or changed. The scripts use `console.log` as their entire output mechanism, which is what `audit-css.mjs` does and why `scripts/**` carries its own eslint config.

## Testing Instructions

Three stages.

1. The scripts run clean and report zero for the markers this repo has genuinely cleared:

```bash
npm run audit:comments
npm run audit:tests
```

Expect exit 0 from both. `audit:tests` should report 0 snapshot tests, 0 skipped, 0 `.only`, and 0 orphans. Those zeroes are the regression guard: if a future edit makes the taxonomy drift back to generic advice, they stop being zero.

2. Spot-check that the findings hold up. Pick any entry from the restating section and read the source. `src/state/worldStore.ts` is the worked example: eight comments that named the function directly below them (`// Add attribute` over `addAttribute`) and are removed in this PR.

3. Full gate:

```bash
npm test && npm run type-check && npm run lint
```

Last run: 393 suites and 2681 tests passed, `tsc` exit 0, lint 0 errors with 143 pre-existing `any` warnings.

## Checklist

File size limits: see Implementation Notes. Accessibility is N/A here, no UI.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,8 @@
     "lint:css": "stylelint \"**/*.css\"",
     "lint:css:fix": "stylelint \"**/*.css\" --fix",
     "audit:css": "node scripts/audit-css.mjs",
+    "audit:comments": "node scripts/audit-comments.mjs",
+    "audit:tests": "node scripts/audit-tests.mjs",
     "docs:check-links": "node scripts/check-docs-links.js",
     "docs:check-symbols": "node scripts/check-docs-symbols.mjs",
     "validate:routes": "node scripts/validate-routes.js",
--- a/scripts/audit-comments.mjs
+++ b/scripts/audit-comments.mjs
@@ -116,7 +116,11 @@ const PATTERNS = {
   // Past-state phrasing only. Bare "used to" is usually present tense
   // ("the salt used to encrypt"), so it is matched with a past-state verb.
   archaeological:
-    /\b(extracted (?:out )?from|moved (?:out )?from|refactored out|refactored from|formerly|previously|renamed from|split out (?:of|from)|used to (?:be|live|call|gate|have|exist|return|do|get|point|sit)|(?:was|were|had) (?:previously|originally)|once (?:was|lived|called)|as of \d)\b/i,
+    /\b((?:was|were|been) extracted (?:out )?from|extracted out from|moved (?:out )?from|refactored out|refactored from|formerly|previously|renamed from|split out (?:of|from)|used to (?:be|live|call|gate|have|exist|return|do|get|point|sit)|(?:was|were|had) (?:previously|originally)|once (?:was|lived|called)|as of \d)\b/i,
+  // "a previously selected skill", "any previously encrypted data": here
+  // "previously" is an adjective describing current data, not a note about how
+  // the code used to be written. An article in front is the tell.
+  adjectivalPrevious: /\b(a|an|any|the|some|each|every|all) previously\b/i,
   // Only line comments draw banners. Block-comment continuation lines are
   // excluded because markdown inside JSDoc ("* - **Bold**:", "* ### Heading")
   // trips any charset loose enough to catch a real divider.
@@ -200,7 +204,10 @@ function scanFile(file) {
       else findings.issueCitation.push(entry);
     }
 
-    if (PATTERNS.archaeological.test(line)) {
+    if (
+      PATTERNS.archaeological.test(line) &&
+      !PATTERNS.adjectivalPrevious.test(line)
+    ) {
       findings.archaeological.push({ at, text: trimmed });
     }
 
@@ -252,6 +259,12 @@ function scanFile(file) {
     // the next line already spells out. "// Add attribute" over "addAttribute:"
     // is the shape this is looking for.
     if (overlapRatio(trimmed, next.line, 4) >= 0.8) {
+      // A comment that matches the line after next as well is heading a GROUP --
+      // a run of sibling constants or literals, like "// Sci-fi" over four
+      // scifi-* preset entries. That labels the run and has to survive; only a
+      // comment that restates one single line is noise.
+      const after = nextCodeLine(lines, next.index + 1);
+      if (after && overlapRatio(trimmed, after.line, 4) >= 0.8) return;
       findings.restating.push({
         at,
         text: trimmed,
--- a/scripts/audit-comments.mjs
+++ b/scripts/audit-comments.mjs
@@ -1,0 +1,357 @@
+/**
+ * Comment audit.
+ *
+ * Lists code comments that the repo's comment hygiene rules say shouldn't exist:
+ * ones that restate the line below them, cite an issue number, narrate refactor
+ * history, or draw a section banner. It is an AUDIT, not a fix: it writes
+ * nothing, deletes nothing, and never fails CI (exit 0 on findings).
+ *
+ * Why this approach (chosen over alternatives):
+ *  - A line-oriented scan with a tiny comment-state machine is enough here. The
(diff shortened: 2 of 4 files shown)


jerseycheese:
All four findings verified against the source before touching anything, and all four held up. Fixed in `6e0dad9b`.

**Jest roots.** Confirmed: `jest.config.cjs` matches `__tests__/**` and `scripts/**` on top of `src`, and there are exactly seven test files there. `--dir` now takes a comma-separated list and defaults to all three roots. `--dir scripts` alone reports 6 files and 82 cases, none of which the audit could previously see.

**Multiline `it.each`.** Confirmed in `surfaceMode.test.ts` at lines 5, 16 and 41, where the table opens on its own line and the case name lands on the `])(` line. Added a lookahead for that closing line, and the body scan now starts from the name rather than the table so brace counting stays correct. Case count goes 2482 to 2580.

**Named assertion helpers.** Confirmed: `\bassert\b` cannot match `assertChoicesVisible` because the boundary fails against the following `C`. `ChoiceSelector.test.tsx:138` was the false positive you identified. Now matches `assert\w*\(`. `noAssertion` drops from 3 to 0, and all three entries were false positives, which also means the tier-1 total this PR's description quoted was made up entirely of them. Description corrected.

**Call-only detection by shape.** This was the best of the four. Requiring the subject to look mock-named made the check depend on what someone called a variable, and your example is exactly right: the three cases at `page.test.tsx:56-96` assert only `expect(notFound).toHaveBeenCalled()` and were missed because the spy is not named `mockNotFound`. Dropped the subject requirement entirely, since a jest call matcher only works on a mock or spy and therefore already establishes it. `mockOnly` goes 173 to 274 and name-mismatch 21 to 31.

The zeroes that exist to catch taxonomy drift are unchanged: 0 snapshot tests, 0 skipped, 0 leaked `.only`.

One clarification rather than a fix: duplicate names are deliberately scoped per describe block, not per file. Jest reports the full `describe > it` path, so the same `it` text under two different describes is a distinct test and usually intentional. Scoped that way the count is 0; a raw per-file grep reports 14, and that gap is the reason for the narrower rule.

Issue #1822, opened 2026-08-16
The world never acts on its own: give the story a clock and off-screen actors


jerseycheese:
## Plain Language Summary

Nothing in the story ever happens on its own. Every event is a direct response to what the player just did, so if the player stops pushing, the story stops. Three 30-turn sessions in two different worlds all confirmed it, and it is the single biggest reason the game stops being engaging after about ten turns.

## Description

This is the root cause behind most of the findings in #1818. Three blind readers scored three separate transcripts with no contact between them, and each named it in their own words:

- Run 1: "Failure means the chosen action does not happen."
- Run 2: "The world only moves when the player moves it. The moment the player walks away from the other characters the game has nothing left to generate but the weather."
- Run 3: counted **five instances of the world acting on its own initiative across 30 turns**, one of which was a repeat of an earlier one and one of which read as a bug.

There is no world simulation, no off-screen actor, and no clock. The generator renders a reaction to the player's action and nothing else.

### Why run 3 is the clean proof

Camp Crystal Lake can always reach for the killer, so a quiet stretch there can be blamed on the model not using an available tool. Harrowgate Mills was built specifically to remove that excuse: a small-town council argument over selling a mill, no threat, nobody can die, every stake social.

Result: **30 turns in one meeting room.** The meeting never formally opens. Turn 9 still offers "open the floor for initial statements" as an available action, and by turn 30 nobody has presented anything. Zero items gained or lost. One location string for all 30 segments. A named deadline arrives at turn 16 and never moves or presses.

The most telling detail across the whole campaign: the highest-scoring ten-turn block of any run is run 3's middle block, and it is the block where an NPC walked in unbidden.

### Consequences the story loaded and then dropped

Run 2's judge listed three the game set up itself and never spent:

- The player stomps out the camp's only fire at turn 17. It never gets dark for anyone.
- The player burns another character's book at turn 2. She never raises it again.
- A character stares at something in the woods for four turns. It never arrives.

Any one of those coming due would have made the final third of that session impossible to write the way it was written.

## Suggested direction

Give the world its own turn. Every turn, something the player is not doing should change: an off-screen actor advances one step, a threat closes one increment, a deadline ticks, or a consequence the player already created comes due.

[...]

PR #1895, opened 2026-08-20
Retry the world clock seed when it comes back empty

### LINKED ISSUE #1873: [BUG] World clock seed can return zero threads on an established session and never retries
## What happened

Camp Crystal Lake, round 5 (#1818): 15 turns played with the world clock off, server restarted with the flag on, session resumed. At turn 16 the orchestrator ran the seed (`hasSessionLedger` went false -> true, the turn's stamp says `open: 0`), but the extraction returned no `opened` threads. The session was then marked seeded, so the seed never ran again; the empty-ledger fallback rule carried turn 17, which opened one thread from prose, and that was the ledger for the remaining 29 turns.

The world description hands the seed four pressure sources on a plate: something watching from the woods, a dead radio, forty minutes of dirt road to town, kids arriving. None became a thread.

## Why (partly a guess, one call to check)

Two candidate causes, both in `src/lib/ai/worldThreadExtraction.ts`:

1. The seed section asks for 3-5 `deadline` or `actor` threads and in the same prompt the ADVANCE/OPEN rules say "OPEN a thread only when THIS segment's prose loads one" and "if the observable-changes line shows nothing changed, be conservative". On an established session the seed and the per-turn rules sit side by side and the conservative rule can win.
2. The seed's kind allowlist is `deadline` or `actor`; a horror threat that is unnamed and unseen may not read as either to the model, and `consequence` is not offered at seed time.

Verify by pulling the turn-16 extraction request and response from the DevTools request log before touching the prompt.

## Fix shape

- Seed rules should say plainly that on the seed turn the "only from this segment's prose" restriction does not apply, and allow `consequence` at seed.
- Do not record the session as seeded when the seed returned zero threads; let the next turn retry (cap at 3 attempts so a world with genuinely nothing to seed does not pay for a retry every turn).
- Unit test: an extraction result with `opened: []` on the seed turn leaves `hasSessionLedger` false.

Related: #1822 (parent), the come-due follow-up filed alongside this one, #1818 round 5.

jerseycheese:
## Description

The world clock's ledger seed could come back with nothing and still mark the session seeded, after which it never ran again. Camp Crystal Lake round 5 hit exactly that: the seed ran at turn 16, opened zero threads, and the empty-ledger fallback carried the remaining 29 turns even though the world description was handing the seed four pressure sources on a plate.

Both suspected causes reproduced in the current code, so the diagnosis in the issue still holds. Two halves to the fix:

**The prompt (`src/lib/ai/worldThreadExtraction.ts:48`).** The seed block asked for 3-5 threads while the per-turn rules two paragraphs above said "OPEN a thread only when THIS segment's prose loads one (max 2 per segment)" and "if the observable-changes line shows nothing changed, be conservative". On an established session there is no reason the model has to read the seed as the winner. The seed now says outright that the OPEN rules are lifted while seeding — seed threads come from the world's standing pressures rather than this segment's prose, the two-per-segment cap is off, and "be conservative" does not apply. It also offers `consequence` alongside `deadline` and `actor`, and says a threat that is unnamed or unseen is still an `actor`, which is what an unseen horror needed to land as anything at all.

**The retry (`src/state/worldThreadStore.ts:222`).** `applyExtraction` used to write the session key before doing anything else, so an all-empty result marked the ledger seeded. It now counts the attempt instead: a seed that opened nothing leaves the key absent and the next turn seeds again. Capped at three attempts (`MAX_SEED_ATTEMPTS`) so a world with genuinely nothing to seed stops carrying the seed block every turn. The counter is a memory-only store field, deliberately outside `partialize` — it is a per-run backstop, not ledger data worth persisting. `clearSessionThreads` drops it alongside the key so a fresh session gets its attempts back.

The decision keys off what was actually opened, not what the model claimed, so a seed whose entries all fail validation retries too.

**The migration (second commit, from Codex's P1).** The sessions this fix exists for already have `sessionThreads[sessionId] = []` sitting in IndexedDB, written by the seed that opened nothing — they would read as seeded forever and never see the retry. The store goes to version 2 and the migration drops stored keys with no threads. A key with an empty list can only mean no thread was ever created for that session (resolving a thread keeps its id in the list), so this hands back exactly the affected sessions and nothing else.

## Related Issue

Closes #1873

Parent: #1822.

## Type of Change

- [ ] Documentation update

## TDD Compliance

- [x] All new code is tested
- [x] All tests pass locally

Red first, on the pre-change tree — 4 failures for the right reasons:

```
● buildWorldThreadPromptSection > lifts the per-turn OPEN rules and offers consequence while seeding
  Expected substring: "The OPEN rules above do not apply while seeding"
● worldThreadStore > applyExtraction > a seed that opened nothing leaves the ledger unseeded so the next turn retries
  Expected: false  Received: true
● worldThreadStore > applyExtraction > stops retrying the seed after three empty attempts
  Expected: false  Received: true
● applyWorldClockUpdates > with the flag on, a seed that opened nothing carries the seed into the next turn
  Expected: false  Received: true
```

The store test the issue asked for is the second one. The `applyWorldClockUpdates` test covers the half a store test cannot: that the next turn actually rebuilds the seed context and hands it back to the extractor.

The migration got the same treatment: `migration drops a persisted ledger key that never got a thread` failed on the version-1 identity migration before the bump.

One existing test asserted the old behavior ("an all-empty result still seeds the session ledger") and was rewritten, since that assertion is the bug. Its useful half — an empty result on an already-seeded ledger leaves it seeded — is kept as its own test.

## User Stories Addressed

As a player resuming a session with the world clock on, I want the world to carry its own pressures rather than an empty ledger, so the story keeps moving without me.

## Flow Diagrams

N/A

## Component Development

N/A - no UI in this change.

## Implementation Notes

`WORLD_CLOCK` gates all of this; nothing changes with the flag off.

The prompt half is an AI-behavior change and has not been validated against a live multi-turn run — this is a prompt wording change with a plausible mechanism, not a measured improvement. The retry cap is what makes that acceptable: if the new wording still comes back empty, the session gets two more turns to seed instead of one shot, and worst case it settles where it already was. Measuring whether the reworded seed actually opens threads on an established session belongs to the next playtest round, on the parent epic.

`src/state/worldThreadStore.ts` is 370 lines, over the 300-line guideline. It was 340 before this change; splitting it is out of scope for a complexity:small fix.

## Screenshots

N/A

## Visual Baseline Changes

None.

N/A

N/A

## Quality Checks

- [x] Linting passed
- [x] Type checking passed
- [ ] Security audit passed
- [x] No unhandled promises

Security audit not run - no dependency or auth surface touched.

## Testing Instructions

```
npm test           # 435 suites, 3027 tests, exit 0
npm run type-check # exit 0
npm run lint       # exit 0 (127 pre-existing warnings, 0 errors)
```

Targeted:

```
npx jest src/state/__tests__/worldThreadStore.test.ts \
         src/lib/narrative/__tests__/applyWorldClockUpdates.test.ts \
         src/lib/ai/__tests__/worldThreadExtraction.test.ts
```

In the app, with `WORLD_CLOCK` on: start or resume a session and watch the DevTools world clock section. Before, a seed that returned nothing left "No ledger for this session yet" forever; now the seed block goes back out on the following turns until something opens or three attempts are spent.

## Checklist

File size: `worldThreadStore.ts` was already over the limit before this change, see Implementation Notes. Docs: none describe the seeding rule; the store's own comments carry it.

### DIFF
--- a/src/lib/ai/__tests__/worldThreadExtraction.test.ts
+++ b/src/lib/ai/__tests__/worldThreadExtraction.test.ts
@@ -65,6 +65,18 @@ describe('buildWorldThreadPromptSection', () => {
     expect(section).toContain('Tone instructions: Slow-burn political dread');
     expect(section).toContain('- Find the missing ledger');
   });
+
+  it('lifts the per-turn OPEN rules and offers consequence while seeding', () => {
+    const section = buildWorldThreadPromptSection({
+      openThreads: [],
+      currentTurn: 16,
+      segmentSignals: { location: 'The lake' },
+      seed: { worldDescription: 'Something watches from the woods', activeGoals: [] },
+    });
+
+    expect(section).toContain('The OPEN rules above do not apply while seeding');
+    expect(section).toContain("'consequence'");
+  });
 });
 
 describe('parseWorldThreadExtraction', () => {
--- a/src/lib/ai/worldThreadExtraction.ts
+++ b/src/lib/ai/worldThreadExtraction.ts
@@ -45,7 +45,7 @@ const formatSeed = (seed: WorldThreadSeedContext): string => {
     material.push(`Player's goals:\n${seed.activeGoals.map((goal) => `- ${goal}`).join('\n')}`);
   }
   return `${SEEDING_HEADING}:
-This session's ledger is empty. Derive its starting pressure sources from the world description, tone instructions and the player's goals below as 'deadline' or 'actor' threads (3 to 5), each with a rough dueByTurn where the prose gives a timescale (treat one turn as a few minutes to an hour of story time; 'six weeks' is roughly turn 30). Prefer pressures that can act without the player: a named antagonist, a vote, a storm, a debt.
+This session's ledger is empty. Open 3 to 5 threads on THIS turn from the material below. The OPEN rules above do not apply while seeding: seed threads come from the world's standing pressures rather than from this segment's prose, the two-per-segment cap is lifted, and 'be conservative' does not apply. Use 'deadline', 'actor' or 'consequence', whichever fits the pressure; a threat that is unnamed or unseen is still an 'actor'. Give each a rough dueByTurn where the prose gives a timescale (treat one turn as a few minutes to an hour of story time; 'six weeks' is roughly turn 30). Prefer pressures that can act without the player: a named antagonist, a vote, a storm, a debt, something in the woods.
 ${material.join('\n')}`;
 };
 
--- a/src/lib/narrative/__tests__/applyWorldClockUpdates.test.ts
+++ b/src/lib/narrative/__tests__/applyWorldClockUpdates.test.ts
@@ -29,7 +29,7 @@ describe('applyWorldClockUpdates', () => {
   let worldId: string;
 
   beforeEach(() => {
-    useWorldThreadStore.setState({ threads: {}, entities: {}, sessionThreads: {} });
+    useWorldThreadStore.setState({ threads: {}, entities: {}, sessionThreads: {}, seedAttempts: {} });
     // jest.setup.ts swaps worldStore for its manual mock, which has no
     // setState; seed the one world through the mock's own createWorld.
     (useWorldStore as unknown as { __resetMocks: () => void }).__resetMocks();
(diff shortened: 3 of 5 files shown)


jerseycheese:
Good catch on the P1 — fixed in 2ccf32a1. Store goes to version 2 and the migration drops stored session keys with no threads, so the sessions that already hit the empty seed get handed back to the capped retry instead of reading as seeded forever. An empty list can only mean no thread was ever created for that session (resolving keeps the id in the list), so it targets exactly those and nothing else. Covered by `migration drops a persisted ledger key that never got a thread`.

jerseycheese:
@codex review

PR #1947, opened 2026-08-25
Put lore term definitions in the real page gutter

### LINKED ISSUE #1595: [BUG] Marginalia lane is too narrow to read between 960px and ~1100px
## Plain Language Summary

Lore definitions now sit in the page margin, which is what we wanted. But on smaller desktop windows the margin is too thin to read comfortably — words break mid-syllable to fit.

## Description

#1592 gave the marginalia a real gutter lane, sized as:

```css
--manuscript-marginalia-width: min(
  12.5rem,
  calc(50vw - 21.25rem - var(--space-8) - var(--manuscript-marginalia-outer-inset) - var(--space-1))
);
```

With the tokens resolved (`--space-8: 2rem`, `--space-4: 1rem`, `--space-1: 0.25rem`) that's `min(200px, 50vw - 392px)`:

| Viewport | Lane width |
|---|---|
| 960px (breakpoint floor) | 88px |
| 1005px | 110px |
| 1100px | 158px |
| ~1184px and up | 200px (capped) |

So the lane only reaches its intended 200px from about 1184px up. Below that it narrows linearly and bottoms out at 88px right at the `@media (width >= 960px)` floor. Take off `var(--space-3)` padding on each side and an 88px note leaves roughly 64px of usable text width.

`overflow-wrap: anywhere` keeps it from overflowing, which is the right call for robustness, but at 64px it's breaking ordinary words mid-syllable on most lines. It's legible in the sense that nothing is clipped or off-screen. It isn't pleasant to read.

The visual assertion in `tests/visual/manuscript-regression-assertions.spec.ts` runs at 1005px, so this band is covered for *correctness* — the note is in the margin, nothing overflows — but the assertion has nothing to say about whether the result reads well. That's the expected division of labour; noting it so nobody reads the green check as "this looks fine at 1005px".

## Steps to Reproduce

1. Play until a segment mentions a lore term with a definition.
2. Size the browser window to roughly 1000px wide.
3. Click the term.

## Expected Behavior

The definition is comfortably readable at any width where the gutter layout is active.

## Actual Behavior

Between 960px and about 1100px the note is 88–158px wide and breaks words mid-syllable to fit.

## Screenshots

N/A — reproduces at any width in the band.

## Environment

- Browser: Chromium
- OS: macOS
- Device: Desktop, 960–1100px viewport width
- Branch: `develop`, after #1592

## Additional Context

A few directions, roughly in order of how much they cost:

- **Raise the breakpoint.** Move the gutter layout from `>= 960px` to wherever the lane clears a sensible minimum (~1184px for the full 200px, or ~1100px if 158px is judged acceptable), and let everything below fall back to the mobile bottom sheet. Cheapest fix, and the bottom sheet is a perfectly good treatment.
- **Floor the lane width and let it eat into the prose column.** Keeps marginalia on more screens at the cost of narrowing the manuscript column, which is load-bearing for DS3's look.
- **Reflow below a threshold** — inline the definition under the paragraph rather than beside it. More work, probably the nicest result.

Worth doing after #1543 rather than before. That epic rewrites the DS3 type scale and the manuscript column, both of which are inputs to this calculation — the `21.25rem` term is literally half the current column width, so #1543 will invalidate the arithmetic regardless of what we pick now.

Related: the same expression assumes the manuscript column is centred in the viewport (`50vw`). That holds post-collapse, since the stage is `1fr` with `.manuscript-rail-spacer` hidden, but it would silently misplace the note if a rail ever came back.

### LINKED ISSUE #1596: Marginalia term definitions as true page-gutter marginalia
## Plain Language Summary

Lore term definitions currently render as a small note pulled to the right edge of the prose column, sitting above the paragraph. The nicer version puts them in the actual page margin, beside the text they annotate. A first attempt at that shipped with a real bug and was reverted; this tracks doing it properly.

## Current Feature

`.manuscript-marginalia-definition` (rendered by `TermDefinition.tsx`) uses `align-self: flex-end; width: 200px;` inside `.narrative-segment`'s flex column, at `@media (width >= 1024px)` in `src/styles/manuscript-session.css`. This is the same mechanism `.choice-outcome-callout` already uses. Both render as their own row, right-aligned, above the segment's prose — not beside it, and not in the page's actual right margin.

## Domain

- [ ] World Configuration
- [ ] Character System
- [ ] Decision Tracking System
- [ ] Decision Relevance System
- [x] Narrative Engine
- [ ] Journal System
- [ ] State Management
- [ ] Other: _________

## Enhancement Description

Put the definition in the real page gutter, beside the paragraph that contains the clicked term, rather than in its own row above it.

## Reason for Enhancement

This was attempted in #1592 (originally scoped to #1582) and reverted before merge. Worth capturing what was learned rather than starting from zero next time.

The mechanism worked: widening `.narrative-history-container` past 100% only when it `:has(.manuscript-marginalia-term)`, narrowing `.narrative-history-segments` by the same amount, and clipping horizontally at `.manuscript-overlay-main` gave the note a real lane outside the prose column without introducing a horizontal scrollbar. Verified this arithmetic held (measured note widths matched the calc'd values exactly across several viewport widths).

What broke it: the vertical clamp in `TermDefinition.tsx`'s layout effect capped the note's bottom against its own segment's height, and when the note was taller than the segment (any definition longer than ~2 lines), the tie-break picked pinning the *bottom* over the *top*:

```js
const adjustedTopOffset =
  maximumTopOffset < minimumTopOffset
    ? maximumTopOffset   // <- pins bottom, top runs off-screen
    : Math.min(topOffsetPx, Math.max(minimumTopOffset, maximumTopOffset));
```

That let the note's top run off the top of `.manuscript-overlay-main` (`overflow: hidden`), clipping the category badge, term name, and type label — the exact parts that make it a definition rather than a floating paragraph. Reproduced at 1440px viewport width, not just narrow ones, so it wasn't an edge case — any realistic lore description triggered it. Measured:

| Width | Description | Note height | Clipped above the fold | Term name visible |
|---|---|---|---|---|
| 1440 | long | 239px | 98px | no |
| 1280 | long | 239px | 98px | no |
| 1005 | long | 401px | 260px | no |
| 960 | long | 530px | 389px | no |

The existing Playwright assertion (`Desktop marginalia definition sits in the right margin outside the prose column`) passed the whole time, because it only bounded `definitionTop` from above (`<=  clickedTermTop + 8`), never checked the note's top stayed inside the visible scroller. A geometric assertion that's true even when the name is invisible is the same trap #1587 was written to close, one level up — the convention held, the specific invariant chosen didn't.

Separately, even where it didn't clip, the lane was uncomfortably narrow. The width formula was `min(200px, 50vw - 392px)`, which only reaches 200px above ~1184px viewport width and bottoms out around 88px at the 960px breakpoint floor — leaves ~64px of text after padding, which breaks ordinary words mid-syllable via `overflow-wrap: anywhere`. This was filed separately as #1595; folding it in here since it's a design input to the same lane-width math, not an independent bug.

## Possible Implementation

Salvage the `:has()` width-carving mechanism from #1592 (it worked); replace the vertical placement logic rather than patching its tie-break:

- Fix the clamp to prefer keeping the *top* of the note visible when it can't fit its segment's slot, and decide deliberately what happens to the overflow: extend past the segment (risk: overlapping the next segment's own note), cap with `max-height` + internal scroll on the note, or let it run down into the gutter below the segment boundary.
- Revisit the lane-width formula together with #1543 (the bolder DS3 redesign) rather than before it — that epic rewrites the type scale and the manuscript column width, and `21.25rem` (half the current column width) is a direct input to this calc. Doing this first risks solving for numbers #1543 immediately invalidates.
- Decide whether `.choice-outcome-callout` should ever want the same treatment. #1327 tracks that question for the callout specifically; once one element has working gutter placement, whether the other should share the lane, get its own, or stay in-flow is worth deciding together rather than reopening the geometry twice.
- Whatever ships needs an assertion that bounds the note's top edge against the *visible scroller region*, not just against the clicked term's position — that's the invariant this round's bug actually violated.

## Alternatives Considered

Shipping the clamp with the tie-break flipped and calling it done was considered and rejected for this round — flipping the sign alone still leaves the overflow question (what happens when the note doesn't fit) unanswered, and answering it well is most of the actual work.

## Additional Context

Current shipped state (align-self, in-flow, above prose) is correct and safe — this is a polish enhancement, not a bug. No urgency; sequencing after #1543 is a deliberate recommendation, not a blocker.

Related: #1327 (callout gutter placement, same open question from the other element's side), #1595 (narrow-lane readability, folded into this issue's scope).

jerseycheese:
## Description

Lore term definitions now render in the page's real right gutter, beside the paragraph that names the term, instead of in their own right-aligned row above the prose.

This salvages the width-carve mechanism from #1592, which measured correct, and replaces the vertical placement logic that broke it. The history container widens into the gutter and the segments narrow back by the same amount, so the prose column keeps its position and measure while the note gets a lane inside the Radix ScrollArea Viewport's `overflow-x: hidden`. That clip is the reason a simpler approach doesn't work: Radix sets the overflow inline, and CSS can't override an inline style without `!important`.

Two deliberate departures from the #1592 round. The carve is scoped to `:has(.manuscript-marginalia-definition)` rather than to the term spans, so with nothing open the surface is byte-identical to today. And the lane starts at 1280px instead of 960px, which is where the arithmetic still leaves 14rem of lane.

The clamp is the part that actually broke last time, and it needed a different answer rather than a flipped sign. Clamping the note against its own segment gives it no slot to fit in whenever the definition runs longer than the paragraph, and #1592's tie-break resolved that by pinning the bottom and pushing the category badge and term name off the top of the scroller. This clamps against the visible scroller instead, and caps the note's height at the visible region *first*. Once the note is guaranteed to fit, the top and bottom bounds can't cross, so there is no tie-break left to get wrong. A definition taller than the whole visible story area scrolls inside its own box rather than running under the fold.

#1595 is folded in as the lane-width input the issue asked for. The old formula bottomed out near 88px, leaving about 64px of text and breaking ordinary words mid-syllable. The floor is now 14rem, about 198px of text after padding and border, roughly 30 characters at 13px. Below 1280px the note keeps the in-flow `align-self: flex-end` presentation it has today, and below 1024px it stays a bottom sheet. Both unchanged.

`.choice-outcome-callout` is deliberately untouched, still on `align-self: flex-end; width: 220px`. #1327 tracks whether it should adopt the lane; the measured geometry is posted there.

## Related Issue

Closes #1596
Closes #1595

## Type of Change
- [ ] Documentation update

## TDD Compliance
- [x] All new code is tested
- [x] All tests pass locally

The assertion was written after the CSS, but it was verified the way a test-first one would have been: reverting `TermDefinition.tsx` to #1592's clamp and re-running it puts the note's top at -23px against a scroller top of 71px, and the test goes red on exactly that line. A test that only ever ran green against the fix is how #1592 shipped its bug.

## User Stories Addressed

As a player reading a scene, when I click a lore term I want its definition in the margin beside the paragraph that mentions it, so I can read the annotation without losing my place in the prose or having the story shift under me.

## Flow Diagrams

N/A

## Component Development

Two story fixes went in, but neither adds a story. The `TermDefinition` molecule story rendered the panel in a bare relative box with no segment or scroll-container ancestry, so it could never show gutter placement and would have kept looking correct while the app surface was broken. It now documents what it can actually show and points at the organism story and the geometry assertions for placement. The `NarrativeDisplay` story still described the note as floating right, which was the #1582 bug and has been `align-self` since.

Gutter placement can't be demonstrated in a Storybook canvas at all, since it depends on the play surface's scroller and prose column. Verification for this one is the browser and Playwright, not Storybook.

## Implementation Notes

Every term in the lane arithmetic is a box on the page rather than a guess, and the comment in the CSS spells out where each comes from:

```
lane = clamp(14rem, min(50vw, 46.25rem) - 26rem, 17.5rem)

  min(50vw, 46.25rem)   half the page; .manuscript-viewport-inner is
                        width: min(100%, 1480px), so it stops growing at 740px
  21.25rem              half .manuscript-main-content-inner's 42.5rem column
  1.5rem + 1.5rem       .manuscript-viewport-inner's horizontal padding, pinned
                        at 1.5rem from 1280px up, plus .manuscript-main-content's
  1rem + 0.75rem        gap to the prose, plus slack off the padding edge
                        (the four above sum to the 26rem subtracted)
```

`21.25rem` is still half the manuscript column: `42.5rem` at `manuscript-session.css:3235` last changed in #1526, and DS3 epic #1543 closed 2026-08-10 without touching this file. The issue recommended sequencing after #1543; that gate is satisfied and the numbers still hold.

All four lane rules are scoped to `.narrative-history-container:has(.manuscript-marginalia-definition)`. That matters for the rule that sets `position: absolute`: unscoped it matches every `TermDefinition` on the page, and outside the history container the two custom properties are undefined, so `width` falls back to auto and the `inset` shorthand is invalid at computed-value time. The panels go absolute with no positioned ancestor and pile up. It caught the Storybook molecule stories, which render standalone by design.

Note that `max()` is not in the stylelint allow-list for `manuscript-session.css`, though `min()`, `clamp()`, `calc()` and `minmax()` are. `globals.css` has a different allow-list that does include `max()`. The lane clamp uses `clamp()`.

The issue's history section says the #1592 attempt "was reverted before merge." It wasn't. #1592 merged with the flagged bug (`88c0a4ca`) and #1598 reverted it the next morning (`dbb1e586`); #1596 was filed in between. The issue's description of current state is accurate, its history isn't.

## Screenshots

Before, at 1280px. The note takes its own row above the prose and pushes the whole story down.

Marginalia before, in its own row above the prose at 1280px

After, at 1280px. The note sits in the gutter beside the paragraph containing the term, and the prose hasn't moved.

Marginalia after, in the page gutter at 1280px

At 1480px the lane opens to its 17.5rem cap.

Marginalia in the gutter at 1480px

The pathological case: a definition taller than the whole visible story area. It fills the visible region, keeps its badge and name at the top, and scrolls inside its own box.

An oversized definition capped at the visible scroller and scrolling internally

## Visual Baseline Changes

None. Both Playwright projects were run in full against this branch and every baseline matched: 150 passed in `chromium`, 10 passed in `tutorials`, with `git status` clean afterwards.

That's the part worth flagging, because #1592 churned `choice-outcome-consequences-ds3`, `tutorial-first-play-step00` and `tutorial-first-play-step01`, and its revert churned the identical set. Three scoping choices are what avoided it here: the carve is gated on an open definition rather than on the term spans, `.manuscript-overlay-main` and `.scroll-smooth` are untouched (the widened container stays inside the content box, so there's nothing to clip and no scrollbar to suppress), and `.choice-outcome-callout` keeps its current geometry.

N/A

Measured in Chromium against this worktree's own dev server on port 3881, with the note open on a definition taller than the paragraph it annotates.

| Viewport | Presentation | Note width | Note left | Prose right | Note top / scroller top | Note bottom / scroller bottom | Horizontal overflow |
|---|---|---|---|---|---|---|---|
| 1480 | gutter lane | 280 (17.5rem cap) | 1096 | 1080 | 254 / 74 | 453 / 886 | none |
| 1280 | gutter lane | 224 (14rem floor) | 996 | 980 | 254 / 74 | 473 / 886 | none |
| 1100 | in-flow fallback | 226 | 664 | 890 | 204 / 74 | 423 / 886 | none |
| 1024 | in-flow fallback | 226 | 626 | 852 | 204 / 74 | 423 / 886 | none |
| 960 | bottom sheet | 960 | 0 | 820 | n/a | n/a | none |
| 375 | bottom sheet | 375 | 0 | 356 | n/a | n/a | none |

The prose column measured identical before and after opening the note at both lane widths: `l=400 r=1080` at 1480 and `l=300 r=980` at 1280. `document.documentElement.scrollWidth === clientWidth` at every width, so the lane isn't bought with a horizontal scrollbar, and nothing lands past the right edge at 375px.

Four edge cases, all with the badge and term name inside the scroller:

| Case | Note top | Note bottom | Scroller | Scrolls internally |
|---|---|---|---|---|
| Term near the bottom of the scroller, 1280 | 430 | 649 | 74-886 | no |
| Definition taller than the scroller, 1280 | 212 | 878 | 74-886 | yes |
| Definition taller than the scroller, 1480 | 212 | 878 | 74-886 | yes |
| Short viewport, 1280x560 | 244 | 463 | 64-551 | no |

In the two oversized cases the note is capped at 666px, which is the intersection of `.manuscript-overlay-main` and the Radix viewport rather than the outer scroller alone. Both can clip, so the placement takes the intersection.

- [x] Linting passed
- [x] Type checking passed
- [ ] Security audit passed
- [x] No unhandled promises

`npm test` reported one suite failing to run on a `SIGSEGV` in a Jest worker (`worldAnalyzer.suggestions`, unrelated to anything here); 3124 tests passed and 0 failed. Re-running that suite on its own passes. `npm run type-check`, `npm run lint`, `npm run lint:css` and `npm run deps:validate` are all clean.

## Testing Instructions

1. Start the dev server and open a session in a world that has lore facts for characters, locations or events, so terms render with a dotted underline. `tests/visual/manuscript-regression-assertions.spec.ts` seeds one if you'd rather drive it from Playwright.
2. At 1280px or wider, click a term in a paragraph and check the note lands in the right gutter beside that paragraph, with the prose staying exactly where it was. Scroll the paragraph to the top of the view first and use a long definition; that's the combination that broke last time.
3. Narrow to 1100px and confirm the note falls back to the in-flow presentation above the prose, and to 900px for the bottom sheet.
4. `TZ=UTC npx playwright test tests/visual/manuscript-regression-assertions.spec.ts --project=chromium -g "marginalia"` for the geometry assertions.

## Checklist

The note is focused on mount, so when the oversized case caps its height the keyboard can scroll it without a separate tab stop, and Escape still dismisses with focus returning to the term.


jerseycheese:
Good catch, and it was real. Fixed in `08cded1c`.

The rule that makes the note absolute was the only one in the lane block without a `.narrative-history-container:has(...)` ancestor, so it matched every `TermDefinition` on the page instead of the ones with a lane to go into. It's now scoped like its two siblings, and the segment's `position: relative` rule is scoped the same way for the same reason.

Reproduced before fixing rather than reading the selector. `AllCategories` at 1280px had all three panels at `position: absolute; top: 48; left: 48` with three overlapping pairs and widths shrink-wrapped to 787/783/663px, since `width` fell back to auto and the `inset` shorthand was invalid at computed-value time. After the fix: three `position: relative` panels at tops 48 / 275 / 482, all 226px wide, zero overlaps.

AllCategories story before, three panels stacked on top of each other

AllCategories story after, three separate in-flow panels

There was a second casualty the finding didn't name: `NarrativeDisplay`'s `WithTermDefinitions` story renders standalone too, so it would have gone absolute the same way at >=1280px. It now shows the `align-self` presentation its doc text claims (`position: relative`, `l=1038 r=1264`).

In the app nothing changes, because `NarrativeDisplay` is only ever rendered by `NarrativeHistory`, which always wraps it in `.narrative-history-container`. Confirmed by measurement rather than inference: the play surface is byte-identical to the previous round at 1480, 1280, 1100, 1024, 960 and 375, note rect included, with `position` resolving to absolute / relative / fixed across the three ranges. `npm run lint:css`, `type-check` and `lint` are clean, and both Playwright projects still pass with no baseline churn (150 in `chromium`, 10 in `tutorials`, `git status` clean after).

Issue #1983, opened 2026-08-28
[EPIC] TurnResolver: settle core session state before the next Decision


jerseycheese:
## Plain Language Summary

The turn lifecycle scatters core state mutations across three layers of fire-and-forget async, so the next Decision reads stale inventory, world threads, conditions, and fatal-outcome state. This epic introduces a TurnResolver that settles all core session state before the player's next choices generate.

## Epic Description

The v1.3 and v1.4 prompt experiment campaigns ran 13+ rounds of wording adjustments across five open issues. The conclusion: the prompt-side approach is exhausted. The surviving problems share one root cause.

When a narrative segment generates and commits, multiple store mutations run independently with no coordination:

- `narrativeGenerator.generateSegment()` fires and forgets lore extraction, item acquisition, and item loss
- `narrativeStore.addSegment()` fires and forgets world clock updates, world state thread updates, and fatal-outcome tagging
- `NarrativeController` defers choice generation via `setTimeout(500)`, hoping the mutations have landed

The result: choice prompts read pre-turn state, fatal tags arrive after the ending check, and world thread advancement races the next turn's continuity context.

The fix is a `TurnResolver` module at the boundary where generated prose becomes a committed Turn. Provider output proposes what happened; the session runtime decides what became true. Core state mutations (inventory, conditions, world threads, fatal outcomes) become awaited steps in the resolver pipeline. Choice generation runs after the resolver returns a settled result, not after a timing heuristic.

The architecture reuses existing stores as persistence adapters, serializes turns per session, and avoids event sourcing. Background-only work (lore extraction, item images, journal copies, analytics) stays asynchronous.

## Domain

- [ ] World Configuration
- [ ] Character System
- [ ] Decision Tracking System
- [ ] Decision Relevance System
- [x] Narrative Engine
- [ ] Journal System
- [x] State Management
- [x] Other: Architecture

## Goals

- Settle core session state (inventory, conditions, world threads, scene, fatal outcomes) before the next Decision generates
- Eliminate fire-and-forget race conditions between turn commitment and choice generation
- Provide a read-only SessionSnapshot that prompt projections consume instead of scattered getState() calls
- Serialize turns per session to prevent concurrent mutation
- Preserve the existing store-based persistence without introducing a second session store

## User Stories

- [ ] #1872 - World clock: overdue threads never come due, advance degrades into re-announcement
- [ ] #1882 - A threat that has arrived should cost the player something
[...]

PR #1984, opened 2026-08-28
Introduce TurnResolver to settle core session state before the next Decision (#1983)

jerseycheese:
## Description

The turn lifecycle was split across fire-and-forget work in the generator, narrative store, and controller. The next Decision could read stale inventory, world threads, and fatal-outcome state because those mutations had not settled.

This PR introduces `TurnResolver` between the controller and generator. It serializes controller-driven turns per session, commits the generated segment, awaits core reconciliation, and returns the post-turn snapshot before choice generation starts.

The review follow-up makes that contract honest when something goes wrong. Provider cancellation is enforced at the resolver's pre-commit generation seam, even if downstream generation ignores the abort signal. Reconciliation now returns an explicit `settled` or `partial` status, reports failures from helpers that otherwise fail open, and blocks the next Decision when core state is partial.

### Scope and deferrals

This PR covers controller-driven turns: the initial scene and per-choice segments routed through `NarrativeController`.

- Item-use turns still bypass the resolver. #1985 tracks that work.
- Choice generation still rereads live stores instead of consuming `TurnResult.snapshot`. #1986 tracks that work.

Those gaps mean this PR is one bounded part of #1983, not completion of the epic.

## Related Issue

Part of #1983. Follow-up scope is tracked in #1985 and #1986.

## Type of Change

- [ ] Documentation update

## TDD Compliance

- [x] All new code is tested
- [x] All tests pass locally

The tests hold reconciliation promises pending, exercise the real store boundary, cancel a generator that ignores its signal, and verify that partial results cannot generate choices.

## User Stories Addressed

For controller-driven turns, this establishes causal ordering between segment commitment, core reconciliation, and the next Decision. It also makes partial reconciliation visible instead of presenting an incomplete store revision as settled.

The integrated live matrix and the two deferred paths remain open under #1983, #1985, and #1986.

## Flow Diagrams

Not applicable.

## Component Development

Not applicable. The controller's rendered output is unchanged.

## Implementation Notes

The main modules are:

| File | Lines | Purpose |
| --- | ---: | --- |
| `src/types/turnResolver.types.ts` | 113 | Turn commands, settlement status, result, snapshot, and reconciliation error types |
| `src/lib/narrative/turnResolver.ts` | 618 | Generation boundary, per-session lock, commitment, reconciliation, and snapshot assembly |
| `src/lib/narrative/resolverGuard.ts` | 21 | Call-scoped resolver ownership check |
| `src/lib/narrative/sessionSnapshotAssembler.ts` | 80 | Authoritative snapshot assembly across session stores |
| `src/lib/narrative/__tests__/turnResolver.test.ts` | 672 | Resolver integration coverage |

`NarrativeController` delegates initial and normal turn lifecycle work to the resolver. It generates choices only after a `settled` result. A `partial` result leaves the committed prose visible, sets a non-retryable error, and stops before another Decision can read incomplete state.

`applyWorldClockUpdates`, `applyWorldStateThreadUpdates`, `processAcquiredItems`, and `processLostItems` keep their fail-open behavior for existing callers. An optional error callback now lets the resolver record those failures in `TurnResult.reconciliationErrors`.

Lore extraction, item image generation, journal copies, analytics, and DevTools continuity recording remain asynchronous because they do not define the core settled revision.

## Screenshots

Not applicable. There are no visual changes.

## Visual Baseline Changes

None.

The follow-up addresses the two remaining in-scope blockers from the latest review:

- The timeout no longer races post-commit reconciliation. Resolver-side cancellation prevents an abort-ignoring generator from committing afterward.
- Core reconciliation failures produce `status: 'partial'`, including failures swallowed inside existing helpers, and both controller paths block choice generation for that result.

The item-use and snapshot-consumption review threads describe valid gaps, but they are intentionally separated into #1985 and #1986.

Not applicable. This change does not alter rendered UI behavior.

- [x] Linting passed
- [x] Type checking passed
- [ ] Security audit passed
- [x] No unhandled promises

Knip, dependency validation, dependency baseline checks, the full Jest suite, and the production build also pass locally.

## Testing Instructions

Run the focused settlement and controller coverage first:

```bash
npm test -- src/lib/narrative/__tests__/turnResolver.test.ts \
  src/components/Narrative/__tests__/NarrativeController.generationError.test.tsx \
  src/components/Narrative/__tests__/NarrativeController.contentGate.test.tsx
```

Then run the code-quality gates:

```bash
npm run type-check
npm run lint
npm run knip
npm run deps:validate
npm run deps:check
```

The final local verification passed 452 suites and 3,206 tests:

```bash
npm test
NEXT_PUBLIC_SITE_URL=https://narraitor-six.vercel.app npm run build
```

## Checklist

`turnResolver.ts` and its integration test exceed the 300-line guideline. This PR keeps the initial and normal turn paths beside their shared reconciliation contract; splitting them during this review fix would add unrelated restructuring.

### DIFF
--- a/src/components/Narrative/NarrativeController.tsx
+++ b/src/components/Narrative/NarrativeController.tsx
@@ -512,18 +512,14 @@ export const NarrativeController: React.FC<NarrativeControllerProps> = ({
       setError(null);
       resetStreamingPreview();
 
-      // Race AI generation against a timeout so a slow response falls back
-      // instead of hanging. The abort signal propagates through the resolver
-      // to the underlying generator call.
+      // The abort signal propagates through the resolver to the underlying
+      // generator call. Only the provider round-trip is timed; reconciliation
+      // (store writes + world-cost extraction) runs unraced so a slow
+      // reconciliation doesn't trigger a duplicate fallback segment.
       const generationAbort = new AbortController();
-      const timeoutPromise = new Promise<never>((_, reject) => {
-        generationTimeoutId = setTimeout(() => {
-          generationAbort.abort();
-          reject(
-            new Error(`Initial generation timed out after ${AI_GENERATION_TIMEOUT_MS}ms`)
-          );
-        }, AI_GENERATION_TIMEOUT_MS);
-      });
+      generationTimeoutId = setTimeout(() => {
+        generationAbort.abort();
+      }, AI_GENERATION_TIMEOUT_MS);
 
       const initialCommand: InitialTurnCommand = {
         sessionId,
@@ -534,10 +530,7 @@ export const NarrativeController: React.FC<NarrativeControllerProps> = ({
         onChunk: handleStreamChunk,
       };
 
-      const turnResult = await Promise.race([
-        resolveInitialTurn(initialCommand, narrativeGenerator),
-        timeoutPromise,
-      ]);
+      const turnResult = await resolveInitialTurn(initialCommand, narrativeGenerator);
 
       // Skip if component unmounted during async operation
       if (!mountedRef.current) {
@@ -553,6 +546,13 @@ export const NarrativeController: React.FC<NarrativeControllerProps> = ({
         setSegments(currentSegments);
         setIsLoading(false);
         return;
+      }
+
+      if (turnResult.reconciliationErrors.length > 0) {
+        logger.warn(
+          '[NarrativeController] Initial turn reconciliation partial:',
+          turnResult.reconciliationErrors.map((e) => e.step)
+        );
       }
 
       const gatedSegment = turnResult.segment;
@@ -576,10 +576,7 @@ export const NarrativeController: React.FC<NarrativeControllerProps> = ({
         generatePlayerChoices();
       }
     } catch {
-      // Error generating initial narrative (timeout, network, etc.)
-      // The resolver may have already committed a segment before the
-      // timeout race rejected. Check the store before inserting a
-      // fallback to avoid duplicates.
+      // Error generating initial narrative (timeout, abort, network, etc.).
       try {
         const alreadyCommitted = getSessionSegments(sessionId);
         if (alreadyCommitted.length > 0) {
@@ -716,17 +713,13 @@ export const NarrativeController: React.FC<NarrativeControllerProps> = ({
       const pacingEscalationRequested =
         !hasWorldClock && isPacingStale(computeTurnsSinceComplication(segments));
 
-      // Race AI generation against a timeout so a slow response falls back
-      // instead of hanging.
+      // The abort signal times out the provider round-trip; reconciliation
+      // runs unraced so a slow side-effect pass doesn't trigger a spurious
+      // error + Retry while the segment was already committed.
       const segmentAbort = new AbortController();
-      const segmentTimeoutPromise = new Promise<never>((_, reject) => {
-        segmentTimeoutId = setTimeout(() => {
-          segmentAbort.abort();
-          reject(
-            new Error(`Segment generation timed out after ${AI_GENERATION_TIMEOUT_MS}ms`)
-          );
-        }, AI_GENERATION_TIMEOUT_MS);
-      });
+      segmentTimeoutId = setTimeout(() => {
+        segmentAbort.abort();
+      }, AI_GENERATION_TIMEOUT_MS);
 
       const command: TurnCommand = {
         sessionId,
@@ -754,14 +747,18 @@ export const NarrativeController: React.FC<NarrativeControllerProps> = ({
         onChunk: handleStreamChunk,
       };
 
-      const turnResult = await Promise.race([
-        resolveTurn(command, narrativeGenerator),
-        segmentTimeoutPromise,
-      ]);
+      const turnResult = await resolveTurn(command, narrativeGenerator);
 
       // Skip if component unmounted during async operation
       if (!mountedRef.current) {
         return;
+      }
+
+      if (turnResult.reconciliationErrors.length > 0) {
+        logger.warn(
+          '[NarrativeController] Turn reconciliation partial:',
+          turnResult.reconciliationErrors.map((e) => e.step)
+        );
       }
 
       const gatedSegment = turnResult.segment;
--- a/src/components/Narrative/NarrativeController.tsx
+++ b/src/components/Narrative/NarrativeController.tsx
@@ -576,42 +576,53 @@ export const NarrativeController: React.FC<NarrativeControllerProps> = ({
         generatePlayerChoices();
       }
     } catch {
-      // Error generating initial narrative — create a graceful fallback segment
+      // Error generating initial narrative (timeout, network, etc.)
+      // The resolver may have already committed a segment before the
+      // timeout race rejected. Check the store before inserting a
+      // fallback to avoid duplicates.
       try {
-        const now = new Date();
-        const segmentId = `seg-${worldId}-fallback-${Date.now()}`;
-        const fallbackSegment: NarrativeSegment = {
-          id: segmentId,
-          content:
-            'The adventure begins. You find yourself at the edge of a new journey. What will you do next?',
-          type: 'scene',
-          characterIds: [],
-          metadata: {
-            characterIds: [],
-            location: 'Starting Location',
-            tags: ['intro', 'fallback'],
-          },
-          sessionId,
-          worldId,
-          timestamp: now,
-          createdAt: now.toISOString(),
-          updatedAt: now.toISOString(),
-        };
-
-        // Add locally and to the store to unblock the UI
-        const gatedFallback = storeSegmentAndTakeGated(fallbackSegment);
-        setSegments((prev) => [...prev, gatedFallback]);
-
-        // Notify parent so it can progress to choices skeleton + generation
-        if (onNarrativeGenerated) {
-          onNarrativeGenerated(gatedFallback);
-        }
-
-        // Kick off choice generation (will provide AI or fallback choices)
-        if (generateChoices && !isSessionEndingSegment(gatedFallback)) {
-          setTimeout(() => {
+        const alreadyCommitted = getSessionSegments(sessionId);
+        if (alreadyCommitted.length > 0) {
+          // The resolver finished generation but reconciliation pushed
+          // past the timeout. Adopt the committed segment instead.
+          setSegments(alreadyCommitted);
+          if (onNarrativeGenerated) {
+            onNarrativeGenerated(alreadyCommitted[0]);
+          }
+          if (generateChoices && !isSessionEndingSegment(alreadyCommitted[0])) {
             generatePlayerChoices();
-          }, 500);
+          }
+        } else {
+          const now = new Date();
+          const segmentId = `seg-${worldId}-fallback-${Date.now()}`;
+          const fallbackSegment: NarrativeSegment = {
+            id: segmentId,
+            content:
+              'The adventure begins. You find yourself at the edge of a new journey. What will you do next?',
+            type: 'scene',
+            characterIds: [],
+            metadata: {
+              characterIds: [],
+              location: 'Starting Location',
+              tags: ['intro', 'fallback'],
+            },
+            sessionId,
+            worldId,
+            timestamp: now,
+            createdAt: now.toISOString(),
+            updatedAt: now.toISOString(),
+          };
+
+          const gatedFallback = storeSegmentAndTakeGated(fallbackSegment);
+          setSegments((prev) => [...prev, gatedFallback]);
+
+          if (onNarrativeGenerated) {
+            onNarrativeGenerated(gatedFallback);
+          }
+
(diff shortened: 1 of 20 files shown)


src/lib/narrative/turnResolver.ts
```diff
@@ -0,0 +1,437 @@
...
+ * 3. Call the generator (prompt building + Gemini call)
+ * 4. Build and commit the NarrativeSegment
+ * 5. Await core reconciliation (world clock, world state threads, inventory)
+ * 6. Stamp fatal-outcome tags from reconciliation
+ * 7. Assemble post-turn snapshot
+ * 8. Release lock
+ * 9. Fire-and-forget background work (lore extraction handled by generator
+ *    outside the guard, if needed in the future)
+ */
+export function resolveTurn(
```
jerseycheese, line 121: [P1] Route item-use turns through this seam too. `processItemUsage()` still calls `generateSegment()` directly, commits with `addSegment()`, then immediately calls `generatePlayerChoices()` (`src/lib/inventory/itemUsageService.ts:165,297,348`). That path is outside `withTurnLock`, and its clock/thread work remains fire-and-forget, so the next Decision can still read stale state after a real player turn. Wire it through the resolver or narrow the PR's acceptance claim.

src/state/narrativeStore.segments.ts
```diff
@@ -170,6 +171,12 @@ export const createNarrativeSegmentActions = (
       logger.error('[NarrativeStore]', 'Failed to update session narrative count:', error);
     }
 
+    // When the TurnResolver drives this session, it awaits these operations
+    // itself, so skip the fire-and-forget tails to avoid duplicate writers.
+    if (isResolverActive(sessionId)) {
```
jerseycheese: [P1] Scope this guard to the resolver-owned commit. `isResolverActive(sessionId)` suppresses the tails for every segment written to the session while a resolver is active. An item-use, bootstrap, or fallback segment that lands during the generation/reconciliation window will skip its own clock and world-state updates, while the resolver only reconciles the segment it created. A resolver-specific commit method or ownership token would avoid dropping unrelated work.

src/lib/narrative/sessionSnapshotAssembler.ts
```diff
@@ -0,0 +1,73 @@
...
+export function assembleSessionSnapshot(sessionId: EntityID): SessionSnapshot {
+  const narrativeState = useNarrativeStore.getState();
+  const characterState = useCharacterStore.getState();
+  const inventoryState = useInventoryStore.getState();
+  const worldThreadState = useWorldThreadStore.getState();
+  const worldState = useWorldStore.getState();
+  const npcState = useNPCStore.getState();
+  const sessionState = useSessionStore.getState();
+
+  const worldId = sessionState.worldId ?? '';
```
jerseycheese: [P2] Build identity from the requested session, not the active singleton. This pairs `sessionId`'s segments and threads with whichever `worldId` and `characterId` happen to be active. After a session switch, or for `readSnapshot(savedSessionId)`, the result can contain another session's character, inventory, world state, lore, and NPCs. Resolve the IDs from `sessionLifecycle`/`savedSessions`, or pass the command's authoritative IDs into the assembler.

src/lib/narrative/turnResolver.ts
```diff
@@ -0,0 +1,437 @@
...
+        const tags =
+          result.worldCost?.fatal && !currentTags.includes('fatal-outcome')
+            ? [...currentTags, 'fatal-outcome']
+            : currentTags;
+        useNarrativeStore.getState().updateSegment(segmentId, {
+          metadata: { ...current.metadata, ...result, tags },
+        });
+      }
+    }
+  } catch (error) {
```
jerseycheese, line 495: [P2] Do not return a fully settled result after core reconciliation failed. World clock, world-state, acquisition, and loss errors are all swallowed here, and some callees fail open internally too. The resolver then snapshots partial state and lets choice generation continue, so prose can say an item was lost or a thread advanced while the authoritative store did not change. Return an explicit partial/failure status and keep the next Decision blocked, or define a compensating policy that makes the committed result honest.

src/lib/narrative/turnResolver.ts
```diff
@@ -0,0 +1,437 @@
...
+
+    // Synchronous NPC sync (already not fire-and-forget)
+    syncNpcMetadata(worldId, result.metadata.characters);
+
+    // Read the settled segment after reconciliation may have stamped tags
+    const settledSegment =
+      useNarrativeStore.getState().segments[storedSegmentId] ?? storedSegment;
+
+    // Post-turn snapshot: the settled revision
+    const postTurnSnapshot = assembleSessionSnapshot(sessionId);
```
jerseycheese: [P2] Feed this snapshot into the next Decision. `TurnResult.snapshot` is never consumed outside the resolver tests; the controller ignores it and `usePlayerChoices` rebuilds context from scattered store reads. Narrative generation also uses only `segments` from the pre-turn snapshot and reads its remaining context from stores. That leaves the new snapshot as a dead projection instead of the interface #1983 calls for, which is also why Knip reports `readSnapshot` as unused.

src/lib/narrative/__tests__/turnResolver.test.ts
```diff
@@ -0,0 +1,466 @@
...
+      expect(result.segment.sessionId).toBe('session-1');
+      expect(result.snapshot).toBeDefined();
+      expect(result.snapshot.sessionId).toBe('session-1');
+    });
+
+    it('awaits world clock updates instead of fire-and-forget', async () => {
+      const generator = makeMockGenerator();
+      const command = makeCommand();
+
+      await resolveTurn(command, generator);
```
jerseycheese: [P2] Prove the await boundary rather than only invocation. Because this mock is already resolved and the assertion only checks `toHaveBeenCalled`, the test would also pass if the resolver went back to `void applyWorldClockUpdates(...)`. Use a deferred promise, assert `resolveTurn()` stays pending, resolve it, then verify the returned snapshot contains the mutation. The world-thread acceptance case should likewise exercise the real store update instead of mocking the updater wholesale.

jerseycheese (commented):
Recommendation: NEEDS REVISION. The TurnResolver is the right seam, but this branch does not yet establish the single settled revision claimed by #1983. I confirmed the four existing unresolved comments against this head; the inline comments below are additional blockers rather than duplicates. CI is green for unit/integration, type-check, lint, build, and both E2E shards. Knip is still red because `readSnapshot` is unused.

src/components/Narrative/NarrativeController.tsx
```diff
@@ -526,13 +524,18 @@ export const NarrativeController: React.FC<NarrativeControllerProps> = ({
...
+        worldId,
+        characterId: characterId ?? '',
+        generateChoices: generateChoices ?? true,
+        signal: generationAbort.signal,
+        onChunk: handleStreamChunk,
+      };
+
+      const turnResult = await Promise.race([
+        resolveInitialTurn(initialCommand, narrativeGenerator),
         timeoutPromise,
```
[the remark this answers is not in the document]
jerseycheese: The fallback check prevents a second opener, but the timeout still races the whole resolver. In the committed-segment branch, line 593 starts choices while the original resolver is still reconciling. The normal-turn path can still report an error and offer Retry after its segment committed. Put the timeout around provider generation inside the resolver, or await the in-flight resolver after detecting a committed segment.

src/lib/narrative/turnResolver.ts
jerseycheese: The call-scoped flag fixes the unrelated-write suppression, but item use still follows the direct `generateSegment` to `addSegment` to `generatePlayerChoices` path. Its clock and thread reconciliation remains fire-and-forget, so this P1 is still open.

src/lib/narrative/turnResolver.ts
jerseycheese: Recording `reconciliationErrors` does not close this because both controller paths ignore the array and generate choices anyway. `applyWorldClockUpdates` and `applyWorldStateThreadUpdates` also fail open internally, so many failures never reach the array. The returned `TurnResult` still claims a settled core revision when it can be partial.

src/lib/narrative/turnResolver.ts
jerseycheese: This is still open. The latest commit explicitly defers snapshot-to-choice wiring, while both controller paths ignore `TurnResult.snapshot` and `usePlayerChoices` rebuilds context from live store reads. That does not meet #1983's requirement that the next Decision consume the settled revision.

src/lib/narrative/turnResolver.ts
```diff
@@ -0,0 +1,499 @@
...
+    segmentId: storedSegmentId,
+    sessionId,
+    characterId,
+    metadata: result.metadata,
+    isFirstSegment: true,
+  });
+
+  syncNpcMetadata(worldId, result.metadata.characters);
+
+  fireLoreExtraction(result, { worldId, sessionId, characterIds: characterId ? [characterId] : [] });
```
jerseycheese: [P2] Avoid extracting the opening scene twice. `generateInitialScene(..., { resolverManaged: true })` still starts extraction at `narrativeGenerator.ts:455` before it reaches its resolver guard, then this call starts a second extraction after formatting. Besides the extra provider call, the first extraction can canonize the pre-formatted draft. Gate the initial generator tail or skip this resolver call for initial turns.

src/lib/narrative/turnResolver.ts
```diff
@@ -0,0 +1,499 @@
...
+  const currentTurn = countWorldClockTurns(allSegments);
+  const session = useSessionStore.getState();
+  let notes: ReconciledSegmentNotes | null = null;
+
+  try {
+    const result = await applyWorldClockUpdates({
+      segment,
+      sessionId,
+      characterId: metadata.characterIds?.[0],
+      playerCharacterId:
```
jerseycheese: [P1] Keep reconciliation bound to the command's character. The snapshot now uses authoritative IDs, but this cost path rereads the active session and drops the player character if the user switches sessions while generation is in flight. `applyWorldStateThreadUpdates` also prefers `useSessionStore().characterId`, so the old segment can update the newly active character's thread. Pass `characterId` through both reconciliation calls instead of consulting the active singleton.

src/lib/narrative/turnResolver.ts
```diff
@@ -0,0 +1,499 @@
...
+    recordUsage: false,
+  });
+
+  const unattestedSpeakers = (
+    result.metadata.continuity?.remainingIssues ?? []
+  )
+    .filter((issue) => issue.type === 'invented-exchange')
+    .map((issue) => issue.entity);
+
+  void extractStructuredLore(result.content, existingLoreContext, {
```
jerseycheese, line 605: [P2] Preserve player-character filtering in the resolver-owned lore tail. The old extraction passed `playerCharacterName`, which tells the model not to create the player as an NPC and strips that name after parsing. This helper drops it, so resolver-driven normal turns can reintroduce the player as lore. Resolve the character name from the authoritative command and pass it here.

jerseycheese (commented):
Follow-up review on `18fd0e0`: the call-scoped guard, initial-scene lock recheck, fatal handling, authoritative snapshot IDs, and deferred-promise tests are fixed. The remaining core gaps are in the open threads and inline comments. The PR body also needs a truth pass before merge: snapshot wiring is explicitly deferred, `turnResolver.ts` is 499 lines and its test is 586, and the test count and implementation notes are stale.

src/lib/narrative/turnResolver.ts
jerseycheese: This remains open as #1986. The PR body now states that choice generation still reads live stores and does not treat the returned snapshot as consumed. Snapshot-to-choice wiring stays outside this bounded PR.

src/components/Narrative/NarrativeController.tsx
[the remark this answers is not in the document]
jerseycheese: Fixed in 4570143bd. The controller no longer races the whole resolver. The abort signal is enforced around generation inside TurnResolver, so reconciliation runs after the listener is removed; a generator that ignores cancellation still rejects before addSegment. The new test holds that generator pending, aborts it, and confirms the session has no committed segment.

src/lib/narrative/turnResolver.ts
jerseycheese: Fixed in 4570143bd. TurnResult now reports settled or partial explicitly. The fail-open clock, thread, acquisition, and loss helpers report their internal errors to the resolver, and both controller paths block choice generation and surface a non-retryable error for a partial result. The focused tests cover helper propagation plus initial and normal choice blocking.

src/lib/narrative/turnResolver.ts
jerseycheese: This remains open as #1985. The PR body now narrows the settlement guarantee to controller-driven turns and names item-use routing as deferred scope, so this PR no longer claims that path is covered.

Issue #1985, opened 2026-08-28
Route item-use turns through TurnResolver


jerseycheese:
## Plain Language Summary

Using an item still creates its story segment and next choices outside `TurnResolver`. That lets the next Decision start before the item-use turn's world clock and world-thread work has settled.

## Current Feature

`processItemUsage()` in `src/lib/inventory/itemUsageService.ts` consumes the item, calls `NarrativeGenerator.generateSegment()`, commits with `addSegment()`, then calls `generatePlayerChoices()` directly. The segment's core clock and thread reconciliation still runs through `addSegment()` as fire-and-forget work, and the path never enters the resolver's per-session lock.

PR #1984 deliberately limits its settled-turn guarantee to controller-driven turns. This issue closes the item-use gap left by that scope decision.

## Domain

- [ ] World Configuration
- [ ] Character System
- [ ] Decision Tracking System
- [ ] Decision Relevance System
- [x] Narrative Engine
- [ ] Journal System
- [x] State Management
- [x] Other: Inventory System / Architecture

## Enhancement Description

Route item-use turns through the same serialized resolver boundary as controller-driven turns. Item consumption, item-specific generation, segment commitment, awaited core reconciliation, and the handoff to choice generation need one session-scoped sequence.

The existing item behavior should stay intact: quantity changes happen once, failed AI generation can use the current fallback narrative, and significant uses can still create journal entries. The settled resolver result should determine when the next Decision can generate.

Acceptance criteria:

- Item-use generation and commitment run inside the resolver's per-session lock.
- World clock and world-state thread reconciliation finish before item use can generate replacement choices.
- A concurrent controller turn and item-use turn for the same session execute serially without suppressing either turn's side effects.
- Item quantity, fallback narrative, ending-segment, and journal behavior remain consistent with the current path.
- Deterministic tests hold a reconciliation promise pending to prove choice generation cannot start early, and cover one concurrent item-use/controller turn case.

## Reason for Enhancement

#1983 defines a Turn as settled before the next Decision reads it. Item use advances the narrative, commits a segment, and replaces the available choices, so leaving it outside the resolver preserves the exact stale-state race the epic is meant to remove.

## Possible Implementation

[...]

PR #1990, opened 2026-09-01
Route item-use turns through TurnResolver

jerseycheese:
## Description

Item-use turns could replace the current Decision before their world clock and world-state work had finished. This routes them through the same per-session `TurnResolver` lock as controller turns, which keeps consumption, generation, commitment, reconciliation, and replacement-choice persistence in one ordered sequence.

The existing behavior stays intact: one item is consumed, provider failures commit fallback prose, ending and partial turns keep the existing choices, significant uses still create journal entries, and the explicitly used item is excluded once from metadata-driven loss processing.

## Related Issue

Closes #1985

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update

## TDD Compliance

The acceptance cases cover the session lock, deferred reconciliation, concurrent controller and item-use turns, fallback prose, duplicate loss metadata, ending and partial settlement, retained choices, journal creation, and the required command object.

- [x] All new code is tested
- [x] All tests pass locally

## User Stories Addressed

- Item use can't expose a replacement Decision until the turn's core state changes have settled.
- Concurrent item-use and controller turns for one session both commit in order without dropping side effects.
- Item quantity, fallback narrative, ending behavior, and significant-item journal entries remain consistent.

## Flow Diagrams

Not applicable. The issue's serialized flow is covered by deterministic resolver tests.

## Component Development

`InventoryList` passes the authoritative command IDs and disables every Use button while any item use is pending. Its Storybook stories were updated for the required props; no CSS or static layout changed.

- [x] Storybook stories created/updated

## Implementation Notes

`processItemUsage` now requires `{ sessionId, worldId, characterId, itemId }` with no positional compatibility overload. `resolveItemUseTurn` returns a discriminated internal outcome while the service keeps the existing `ItemUsageResult` contract.

Normal, initial, and item-use turns share the same private commitment and reconciliation helper. Item-specific prompt and fallback construction moved to `itemUsageNarrative.ts`, where the injected generator runs with `resolverManaged: true`.

Choice generation and persistence stay inside the item-use lock. Decisions are cleared only after replacement generation succeeds. Partial settlement exposes the same non-retryable pause state used by `NarrativeController`, while journal creation happens after the resolver releases the lock.

`turnResolver.ts` was already over the repository's 300-line guideline and remains over it. The shared lock and private settlement pipeline stay together here; item-specific narrative construction was split into its own 194-line module.

## Screenshots

Not applicable. The UI change is limited to the pending interaction state; layout and styling didn't change.

## Visual Baseline Changes

None.

The pre-PR scope audit found no out-of-scope changes or missing acceptance criteria. Every changed file maps to the resolver path, command-object caller migration, regression coverage, or the existing game-mechanics guide.

Not applicable. The pending-button interaction is covered by a component test and doesn't change the DOM structure or styling.

- [x] Linting passed
- [x] Type checking passed
- [x] Security audit passed
- [x] No unhandled promises

Also passed `npm run knip`, `npm run deps:validate`, and `npm run deps:check`.

## Testing Instructions

Run the focused regression suites first:

```bash
npm test -- src/lib/narrative/__tests__/turnResolver.test.ts src/lib/inventory/__tests__/itemUsageService.test.ts src/lib/inventory/__tests__/itemUsageService.choiceSkip.test.ts src/components/inventory/__tests__/InventoryList.itemUsage.test.tsx src/components/inventory/__tests__/InventoryList.test.tsx --runInBand
```

Then run the full unit suite and static gates:

```bash
npm test -- --maxWorkers=2
npm run type-check
npm run lint
npm run knip
npm run deps:validate
npm run deps:check
```

The default-parallel `npm test` command hit unrelated 5-second UI timeouts on this machine. Each affected case passed twice in isolation, and the full bounded-worker run passed 452 suites and 3,218 tests. Lint completed with zero errors and the repository's existing visual-test warnings. No local visual suite was run because layout and styling didn't change; CI covers the visual and E2E checks.

## Checklist

### DIFF
--- a/src/components/inventory/InventoryList.tsx
+++ b/src/components/inventory/InventoryList.tsx
@@ -4,7 +4,6 @@ import React, { useState } from 'react';
 import { clsx } from 'clsx';
 import { Shield } from 'lucide-react';
 import { useInventoryStore } from '@/state/inventoryStore';
-import { useSessionStore } from '@/state/sessionStore';
 import { EntityID } from '@/types/common.types';
 import { InventoryItem } from '@/types/inventory.types';
 import { Button } from '@/components/ui/button';
@@ -22,6 +21,8 @@ import { DropConfirmationDialog } from './DropConfirmationDialog';
 
 interface InventoryListProps {
   characterId: EntityID;
+  worldId: EntityID;
+  sessionId: EntityID;
   className?: string;
 }
 
@@ -113,6 +114,8 @@ const InventoryItemImage: React.FC<InventoryItemImageProps> = ({
  */
 export const InventoryList: React.FC<InventoryListProps> = ({
   characterId,
+  worldId,
+  sessionId,
   className = '',
 }) => {
   // Select items and character inventory to re-render on changes
@@ -147,7 +150,6 @@ export const InventoryList: React.FC<InventoryListProps> = ({
       .map((id) => itemsObject[id])
       .filter((item): item is InventoryItem => Boolean(item));
   }, [itemsObject, characterInventories, characterId]);
-  const sessionId = useSessionStore((state) => state.id);
   const toggleEquipItem = useInventoryStore((state) => state.toggleEquipItem);
   const [usingItemId, setUsingItemId] = useState<EntityID | null>(null);
   const [errorFeedback, setErrorFeedback] = useState<string | null>(null);
@@ -167,11 +169,12 @@ export const InventoryList: React.FC<InventoryListProps> = ({
     setErrorFeedback(null);
 
     try {
-      const result = await processItemUsage(
+      const result = await processItemUsage({
+        sessionId,
+        worldId,
         characterId,
         itemId,
-        sessionId || undefined
-      );
+      });
 
       if (result.success) {
         // Success path handled via narrative segment entry; no inline feedback needed.
--- a/src/components/inventory/InventoryList.tsx
+++ b/src/components/inventory/InventoryList.tsx
@@ -309,9 +309,7 @@ export const InventoryList: React.FC<InventoryListProps> = ({
                           className="manuscript-inventory-action-button"
                           variant="outline"
                           onClick={() => handleUseItem(item.id)}
-                          disabled={
-                            usingItemId === item.id || item.quantity <= 0
-                          }
+                          disabled={usingItemId !== null || item.quantity <= 0}
                         >
                           {usingItemId === item.id ? 'USING...' : 'USE'}
                         </Button>
--- a/src/lib/inventory/itemUsageService.ts
+++ b/src/lib/inventory/itemUsageService.ts
@@ -1,32 +1,26 @@
-// Item usage service - handles item usage logic, narrative generation, and journal integration
-
-import type { InventoryItem, ItemUsageResult, StandardInventoryCategory } from '@/types/inventory.types';
-import type { EntityID } from '@/types/common.types';
 import { createDefaultGeminiClient } from '@/lib/ai/defaultGeminiClient';
-import { useWorldStore } from '@/state/worldStore';
-import { useCharacterStore } from '@/state/characterStore';
-import { createItemUsageJournalEntry } from './itemUsageJournalIntegration';
 import { NarrativeGenerator } from '@/lib/ai/narrativeGenerator';
+import { PARTIAL_RECONCILIATION_ERROR } from '@/lib/narrative/narrativeErrors';
+import { resolveItemUseTurn } from '@/lib/narrative/turnResolver';
+import { useJournalStore } from '@/state/journalStore';
+import { useNarrativeStore } from '@/state/narrativeStore';
+import type {
+  InventoryItem,
+  ItemUsageResult,
+  StandardInventoryCategory,
+} from '@/types/inventory.types';
 import type { NarrativeGenerationResult } from '@/types/narrative.types';
-import { isSessionEndingSegment } from '@/lib/narrative/isSessionEndingSegment';
-import { mergeTurnTags } from '@/lib/narrative/turnTags';
-import { safeTrim } from '@/lib/utils';
+import type { ItemUseTurnCommand } from '@/types/turnResolver.types';
+import {
+  generateItemUsageNarrative as generateResolverManagedItemNarrative,
+  type ItemUsageNarrativeDetails,
+} from './itemUsageNarrative';
+import { createItemUsageJournalEntry } from './itemUsageJournalIntegration';
 
-import Logger from '@/lib/utils/logger';
-const logger = new Logger('ItemUsageService');
-
-interface ItemUsageNarrativeDetails {
-  wasConsumed?: boolean;
-  remainingQuantity?: number;
-  previousQuantity?: number;
-}
+export { buildUsageNarrative } from './itemUsageNarrative';
 
 /**
- * Determines if an item's usage is narratively significant enough to warrant
- * AI-generated effects and journal entries.
- *
- * Significant items: quest items, equipment, documents, valuables
- * Insignificant items: common consumables, miscellaneous items
+ * Determines whether item usage warrants a journal entry.
  */
 export function isNarrativelySignificant(item: InventoryItem): boolean {
   const significantCategories: StandardInventoryCategory[] = [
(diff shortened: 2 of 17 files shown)


src/lib/narrative/turnResolver.ts
```diff
@@ -580,6 +733,32 @@ async function reconcileCoreSideEffects({
...
+  }
+
+  const excludedName = normalizeText(excludedItem.name, NORM_NAME).toLowerCase();
+  let hasExcludedExplicitUse = false;
+
+  return itemsLost.filter((item) => {
+    const matchesExplicitUse =
+      item.itemId === excludedItem.id ||
+      (!item.itemId &&
+        normalizeText(item.name, NORM_NAME).toLowerCase() === excludedName);
```
jerseycheese: This exclusion is stricter than the thing it's guarding against. `processLostItems` matches through `itemNamesMatch`, which does whole-word containment, then a 15% edit-distance budget, then an AI similarity call. The filter here only accepts an exact `itemId` or an exact normalized name, so most variants the model produces walk straight past it.

Concrete case: 5x "Health Potion", player uses one, `useItem` takes the stack to 4. The model reports `itemsLost: [{ name: 'health potions', quantity: 1 }]` with no itemId. This filter keeps the entry because the names differ, then `namesAreObviousMatch` accepts it (edit distance 1 against a budget of 2) and `removeItem` drops another unit. Two potions gone for one use, which is the double-consumption the filter exists to stop. Using `itemNamesMatch` here, or at minimum the same matching ladder the processor uses, would close it.

src/lib/narrative/turnResolver.ts
```diff
@@ -354,7 +309,200 @@ async function resolveInitialTurnInner(
...
+    segmentId: `seg-${worldId}-item-${itemId}-${Date.now()}`,
+    sessionId,
+    worldId,
+    characterId,
+    isFirstSegment: false,
+    isFatalCriticalFailure: false,
+    playerCharacterName: character.name,
+    reconciliationPolicy: {
+      skipItemAcquisition: true,
+      excludedItemLoss: { id: item.id, name: item.name },
```
jerseycheese: `excludedItemLoss` goes in unconditionally, but `useItem` only consumes when `categoryId === 'consumables' || item.stackable`. For anything else `wasConsumed` comes back false and the item is still sitting in the inventory, so suppressing the model's reported loss for it throws away a real loss.

Player uses a non-stackable quest item, the prose has the key snap off in the lock, the model reports it under `itemsLost`. The filter drops that entry on the id match, `processLostItems` never sees it, and the destroyed key stays usable forever. Gating this on `usage.wasConsumed` should cover it.

jerseycheese (commented):
Five bugs from a careful pass over the diff. The lock plumbing and the shared `commitAndSettleGeneratedTurn` extraction both hold up, and the command-object migration covers the one production caller, so everything below sits in the item-use path.

The two worth fixing before merge are the loss-exclusion filter, which is stricter than the matcher it's guarding against, and the fact that it fires even when nothing was consumed. Either one can leave the inventory wrong after a single use. Details inline.

src/lib/narrative/turnResolver.ts
```diff
@@ -354,7 +309,200 @@ async function resolveInitialTurnInner(
...
+      },
+    };
+  }
+
+  const usageDetails = {
+    wasConsumed: usage.wasConsumed,
+    remainingQuantity: usage.remainingQuantity,
+    previousQuantity: usage.previousQuantity,
+  };
+  const generated = await generateItemUsageNarrative(
```
jerseycheese, line 374: Item-use turns quietly lost narrative loss inference here. `generateSegment` returns at the `isResolverManaged` guard (`narrativeGenerator.ts:208`), which sits above the `inferItemsLostFromNarrative` block at line 273. `resolveTurnInner` re-implements that inference itself, this path doesn't.

Before the change the item-use path called the generator un-managed, so inference ran. Now if the beat reads "you cut the frayed rope loose and it vanishes into the dark" and the model tags nothing under `itemsLost`, the rope stays in the inventory and later prompts keep describing the player carrying it. Either port the same inference block down here, or make it explicit that item-use beats don't get it.

src/lib/inventory/itemUsageService.ts
```diff
@@ -40,343 +34,54 @@ export function isNarrativelySignificant(item: InventoryItem): boolean {
...
 
-  // Use the item via inventory store
-  const usageResult = inventoryStore.useItem(characterId, itemId);
-
-  if (!usageResult.success) {
-    return usageResult;
+  if (outcome.turn.status === 'partial') {
+    useNarrativeStore
+      .getState()
+      .setGenerationError(PARTIAL_RECONCILIATION_ERROR);
```
jerseycheese: Nothing on the item path ever clears this. `clearGenerationError` only gets called from `NarrativeController`, in `generateNextSegment` and `handleRetry`, and the error is `retryable: false` so there's no retry control to hit either.

A partial item-use turn puts up "Stop here rather than making another choice", and then a second item use that settles cleanly leaves the banner sitting there. It only clears when the player makes a story choice, which is the one thing the banner told them not to do.

src/components/inventory/InventoryList.tsx
```diff
@@ -167,11 +169,12 @@ export const InventoryList: React.FC<InventoryListProps> = ({
     setErrorFeedback(null);
 
     try {
-      const result = await processItemUsage(
+      const result = await processItemUsage({
```
jerseycheese, line 172: `usingItemId` (line 168) holds a single id, so starting a second use re-enables the first item's button while its request is still in flight, and the first request's `finally` clears the flag for the second. That was already true before this PR, but item-use turns now queue behind the session lock, so both can be outstanding for the length of a full AI turn instead of overlapping for a moment.

Click Use on A, then on B: A's button reads USE again while A is still going. A resolves, sets the flag to null, and B's button reads USE while B is still queued. Another click there consumes a second unit of B. A set of in-flight ids, or just disabling every Use button while one is pending, would close it.

src/lib/inventory/itemUsageService.ts
jerseycheese: Fixed at the source in c05665dc5. A partial item-use result now installs the terminal pause inside the lock, and queued or later item uses for that active session are rejected before mutation or generation. That means a clean item turn can no longer occur behind the stale banner, and the pause remains visible as intended.

src/components/inventory/InventoryList.tsx
jerseycheese: Fixed in c05665dc5. InventoryList now disables every Use button while any item use is pending. The controlled-promise test clicks a second item during the first request, proves no second service call starts, and verifies the buttons re-enable after settlement.

src/lib/narrative/turnResolver.ts
```diff
@@ -580,6 +733,32 @@ async function reconcileCoreSideEffects({
...
+
+  return itemsLost.filter((item) => {
+    const matchesExplicitUse =
+      item.itemId === excludedItem.id ||
+      (!item.itemId &&
+        normalizeText(item.name, NORM_NAME).toLowerCase() === excludedName);
+
+    if (!hasExcludedExplicitUse && matchesExplicitUse) {
+      hasExcludedExplicitUse = true;
+      return false;
```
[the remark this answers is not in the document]
jerseycheese: I'm keeping the full exclusion here. The issue contract says the selected item is consumed exactly once and excluded from metadata-driven loss processing; decrementing the same metadata record would let one item-use turn consume additional units. The quantity-2 test now pins that behavior while unrelated losses still reconcile.

src/lib/narrative/turnResolver.ts
```diff
@@ -354,7 +309,200 @@ async function resolveInitialTurnInner(
...
+    isFatalCriticalFailure: false,
+    playerCharacterName: character.name,
+    reconciliationPolicy: {
+      skipItemAcquisition: true,
+      excludedItemLoss: { id: item.id, name: item.name },
+    },
+  });
+
+  if (turn.status === 'settled' && !turn.isEnding) {
+    await replaceChoicesAfterItemUse(command, item.id, turn, generator);
```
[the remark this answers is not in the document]
jerseycheese: Fixed in c05665dc5. The resolver now records the terminal partial pause before releasing the session lock and rejects queued or later item-use turns for the active paused session before inventory mutation or generation. Tests cover a queued second use and preserve the existing decisions and pause error.

src/lib/narrative/turnResolver.ts
jerseycheese: Fixed in c05665dc5. The exclusion filter now delegates name comparisons to itemNamesMatch, the same matcher used by loss processing. The new pluralized health potions test proves the explicit use only removes one unit.

src/lib/narrative/turnResolver.ts
jerseycheese: Fixed in c05665dc5. excludedItemLoss is now set only when usage.wasConsumed is true. A non-consumed quest item reported as destroyed now reaches processLostItems, with a regression test covering that path.

src/lib/narrative/turnResolver.ts
jerseycheese: Fixed in c05665dc5. I extracted the existing prose-loss inference into a shared helper and run both normal and resolver-managed item-use results through it before settlement. The regression test uses the real inference against an Old Key loss in the generated prose.

Issue #1988, opened 2026-08-31
Weekly code-health sweep — 2026-08-31


jerseycheese:
## Plain Language Summary

Five of the seven feature flags have no line in `.env.example`, including two the code calls kill switches — so the switches exist but nobody copying the example file can find them.

## Context

Automated weekly code-health sweep, report-only — no code changed, no branch, no PR. Ran against a throwaway clone of `develop` @ `b70a4083`, 21 commits on from last week's `2861285`. Lanes run: `knip` in both modes, `npm run lint`, `type-check`, `stylelint`, `lint:layout-usage`, `lint:ds-canon`, `audit:css`, `audit:comments`, `audit:tests`, `skott:check`, `docs:check-links`, `docs:check-symbols`, `validate:routes`, `deps:check`, plus a hand pass over the three modules added this week and a duplicate-definition scan across all 705 production source files.

**Every tool lane is green, and this week two of them went quiet for the first time.** `knip` now reports nothing in *either* mode — the one file `knip --production` flagged last week is gone. `skott:check` reports 1 cycle against a baseline of 1, down from 6; the five barrel-import cycles were real and were fixed. `deps:check` holds at 12 suppressed violations, down from 17. `npm run lint` exits 0 with 128 warnings, every one of them the deliberate `no-explicit-any` under `tests/visual/` and zero in `src/`. `type-check`, `stylelint`, `lint:layout-usage`, `lint:ds-canon`, `docs:check-links`, `validate:routes` and `audit:css` all exit 0, and `audit:css` holds at 7 potentially-unused selectors, all known false positives.

So the tools have nothing left to say, and every finding below came from reading. Last week's sweep #1951 landed hard: six of its seven findings are fixed and verified (details at the bottom). The seventh is now #1987.

Ranked by value-to-effort. Eight findings, nothing truncated.

## Findings

[...]

Issue #1987, opened 2026-08-31
Two live Character interfaces with the same name and different shapes


jerseycheese:
## Plain Language Summary

Two different types in this codebase are both called `Character`, they describe the same thing differently, and which one you get depends on which file you happen to be reading.

## Context

Carried forward from the weekly code-health sweep #1951, where it was finding 2. Six of that sweep's seven findings were fixed by #1964 and #1955; this is the one that survived. It gets its own issue rather than another checkbox on a snapshot, because a checklist item nobody actions just gets re-filed next week.

Re-verified against `develop` @ `b70a4083` on 2026-08-31, and the description below is corrected from #1951's — the original overstated the cost, and the fix for the rest of that sweep is the reason why.

**What #1951 said, and what changed.** The original finding argued the duplication was already being paid for in code: roughly 70 lines of synthetic character construction across two portrait call sites, plus a 35-line `adaptStoreCharacterToUtil` that invented `experience: 0`, empty inventories and `isActive: true` just to bridge the two shapes. **All of that is gone.** #1964 narrowed `buildPortraitPrompt` to a `PortraitSubject` and `evaluateSkillCheck` to a `SkillCheckSubject`, which deleted both synthetic blocks and shrank the adapter to a 12-line `toSkillCheckSubject` (`src/lib/narrative/evaluateDecisionSkillChecks.ts:34`) that picks two arrays and fabricates nothing. So the concrete, countable cost the original finding leaned on no longer exists.

What remains is smaller and harder to measure: the naming collision itself.

## Findings

- [x] **There are two live `Character` interfaces with the same name and different shapes.** `src/types/character.types.ts:14` and `src/state/characterStore.types.ts:50` — **medium**. The store's `CharacterAttribute` is `{id, characterId, worldAttributeId?, name, baseValue, modifiedValue, category?}`; the types one is `{attributeId, value}`. The store has `level`, `alignment` and `isPlayer`; the types one has none of them. The store types `background.relationships` as `unknown[]`; the types one has a real `CharacterRelationship`. 17 files import from `character.types`, 10 from `characterStore.types`, and `src/types/index.ts:13` re-exports the types-file one as the app's public `Character` — the one the store does not use.

[...]

PR #2006, opened 2026-09-02
Give the duplicated helpers one home and document every feature flag

jerseycheese:
## Description

Clears seven of the eight findings from the 2026-08-31 code-health sweep. Two things are going on.

The flag half is the only finding with a real failure mode. `.env.example` documented 2 of the 7 feature flags while `featureFlags.ts` claimed in its docblock that the example file restated none of them, which meant `WORLD_CLOCK` and `UNRECORDED_EXCHANGE_GUARD` - both shipped, both on, both described in code as the kill switch for their feature - were only findable by reading `src/lib/`. All seven flags now have a commented override line, the docblock says what's actually true, and a test asserts the two files agree so the next flag added can't quietly skip one.

The rest is the same shape seen four times: a small helper defined two or three places under one name, so a grep comes back with two answers and you have to read both to work out which one is live. None of them was a bug and none was going to become one; the value is that the grep now returns one answer. Kept deliberately mechanical - no behavior changes, no opportunistic refactoring.

| Finding | Change |
|---|---|
| 1 | All 7 flags documented in `.env.example`; `featureFlags.ts` docblock corrected |
| 2 | `escapeRegExp` moved to `lib/utils/formatters.ts`; three copies deleted, one of which was redeclared inside a function body |
| 3 | `textFormatter.ts` composes the one `normalizeWhitespace` in `textNormalization.ts` instead of keeping a private near-twin |
| 4 | Ending-image route takes `characterName` instead of a whole `Character` |
| 5 | **Not in this PR** - see below |
| 6 | `termPattern` exported from `continuityLedger.ts`, byte-identical twin deleted from `continuityGuardrail.ts` |
| 7 | `ImageRequestCoordinator`'s never-used constructor option deleted |
| 8 | Checkpoint generator's private `sanitizeString` renamed `trimmedOr` |

**Finding 5 (the `NarrativeController` extraction) is deliberately out of scope here.** It is the one real piece of work on the list rather than a quick win, and pulling a 269-line function out of a 993-line component does not belong in a sweep PR. Its checkbox stays unticked on the issue.

## Related Issue

Closes #1988

## Type of Change

- [x] Documentation update

## TDD Compliance

- [x] All new code is tested
- [x] All tests pass locally

Not strictly test-first. Six of the seven findings move existing code without changing what it does, so the existing suites are the regression net and they were the check at each step. The one new guard (the `.env.example` parity test) was written after the flag lines went in, but it was verified against the pre-fix file to make sure it actually fails: run against `HEAD:.env.example` it reports exactly the five undocumented flags, and against the fixed file it reports none.

## User Stories Addressed

None directly. This is maintenance. The nearest thing to a user story is the operator one behind finding 1: someone needs to turn off the world clock during an incident and should be able to find the switch in `.env.example` rather than by reading `src/lib/featureFlags.ts`.

## Flow Diagrams

Not applicable.

## Component Development

No new components and nothing rendered changed. The two component files touched (`FormattedNarrativeContent.tsx`, `EndingScreen.tsx`) only lost a duplicated helper and swapped a request field; no JSX, CSS, or class names moved.

## Implementation Notes

`escapeRegExp` went into `lib/utils/formatters.ts` rather than `lib/utils/typeGuards/`. The sweep pointed at last week's typeGuards move as the precedent, and the principle carries - one home under `lib/utils`, exported through the barrel - but a regex escaper is not a type guard, and `formatters.ts` is already where the generic string helpers live next to `safeTrim`. The lore modules import it from `@/lib/utils/formatters` directly rather than through the barrel, to keep them off the module graph that `skott` watches.

Finding 3 needed a judgment call, so worth spelling out. The two `normalizeWhitespace` bodies diverge on newline handling: the `textFormatter` one leaves newlines alone, and the `textNormalization` one either collapses `\n{3,}` to a paragraph break or flattens everything. `formatAIResponse` now calls `normalizeText` with structure preserved and quote/special-character rewriting switched off, which is behavior-identical, because `formatParagraphs` runs that same `\n{3,}` collapse two lines later anyway. Quote rewriting had to stay off - `formatDialogue` inserts quotes of its own further down the pipeline.

Finding 4 changes an internal client/server contract with no external consumers, so both sides moved together with no back-compat shim: the route's `GenerateEndingImageRequest`, the `endingImageApi` client, and the two callers (`EndingScreen`, `EndingImageDebugSection`). #1987 lists the same item as its low finding; it is fixed here, not there.

## Screenshots

Not applicable, nothing visual changed.

## Visual Baseline Changes

None. This PR doesn't touch `tests/visual/**-snapshots/**`.

Ran the repo's `narraitor-pattern-alignment-skill` over the diff before committing and `narraitor-architecture` before deciding where `escapeRegExp` should live. Nothing flagged: no new domain-boundary crossings, no design-token or styling surface touched, no new exports beyond the two the dedup needs (`escapeRegExp`, `termPattern`), both of which have real consumers.

Not applicable. No runtime or rendered behavior changed, so there was nothing a browser pass would show that the unit suites don't.

- [x] Linting passed
- [x] Type checking passed
- [ ] Security audit passed
- [x] No unhandled promises

Security audit not run; nothing here touches dependencies, secrets, or the provider-key path.

## Testing Instructions

Everything here is covered by the unit suites, so the check is the gate rather than a click-through:

```
npm test            # 452 suites, 3265 tests, all pass
npm run type-check  # exit 0
npm run lint        # exit 0, 128 warnings, all pre-existing under tests/visual
npm run knip        # exit 0
npm run deps:validate  # no violations, baseline holds at 10
npm run deps:check     # baseline holds, all 10 still reproducing
```

`npm run lint:css` was skipped because no `.css` file is touched. The build and Storybook build are left to CI.

For the two findings worth eyeballing: `grep -rn "escapeRegExp\|normalizeWhitespace" src/` should now show one definition each, and `grep NEXT_PUBLIC_FEATURE .env.example` should list all seven flags.

## Checklist

On the file-size line: no file crosses 300 lines because of this PR. `formatters.ts` (495), `continuityGuardrail.ts` (509) and `continuityLedger.ts` (336) were already over, and the net movement is downward everywhere except `formatters.ts`, which gains 21 lines as the new home for `escapeRegExp` and loses that many again across the three files it replaces. Accessibility is unaffected, no markup changed.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -8,8 +8,16 @@ GEMINI_API_KEY=your-gemini-api-key-here
 # Feature flags. Their defaults live in src/lib/featureFlags.ts - leave these
 # unset to get them. Each line below is the override, not the default, so
 # uncommenting one flips that flag away from what the code ships with.
+# Every flag gets a line here, including the two kill switches for shipped
+# features (WORLD_CLOCK, UNRECORDED_EXCHANGE_GUARD) - you shouldn't have to
+# read src/lib/ to find those during an incident.
 # NEXT_PUBLIC_FEATURE_BUFFERED_STREAMING=true
 # NEXT_PUBLIC_FEATURE_PROGRESSIVE_DISCLOSURE=false
+# NEXT_PUBLIC_FEATURE_WORLD_CLOCK=false
+# NEXT_PUBLIC_FEATURE_WORLD_COST=true
+# NEXT_PUBLIC_FEATURE_WORLD_DESCRIPTION_IN_SCENE=false
+# NEXT_PUBLIC_FEATURE_UNRECORDED_EXCHANGE_GUARD=false
+# NEXT_PUBLIC_FEATURE_SETTLED_COMMITMENT_CHOICES=true
 
 # Public site origin, used for metadataBase, robots.txt, and the sitemap.
 # Leave unset for local development - getSiteUrl() falls back to
--- a/src/app/api/generate-ending-image/__tests__/route.test.ts
+++ b/src/app/api/generate-ending-image/__tests__/route.test.ts
@@ -149,6 +149,33 @@ describe('/api/generate-ending-image', () => {
     expect(mockGenerate).not.toHaveBeenCalled();
   });
 
+  // CONTEXT.md rules "hero" out of the vocabulary: a world can be a slasher or a
+  // civic drama, and this prompt steers the image model. "hero banner" further
+  // down is the layout term and is meant to stay.
+  it('puts the character in the scene rather than casting them as a hero', async () => {
+    await POST(
+      makeRequest({
+        ending: { ...mockEnding, tone: 'triumphant' },
+        world: { ...mockWorld, name: 'Camp Crystal Lake', genre: 'slasher' },
+        characterName: 'Sarah',
+      })
+    );
+
+    const prompt = mockClient.generateContent.mock.calls[0][0] as string;
+    expect(prompt).toContain('Sarah standing tall');
+    expect(prompt).not.toMatch(/\bthe hero\b/i);
+  });
+
+  it('falls back to the character when no name is given', async () => {
+    await POST(
+      makeRequest({ ending: { ...mockEnding, tone: 'triumphant' }, world: mockWorld })
+    );
+
+    const prompt = mockClient.generateContent.mock.calls[0][0] as string;
+    expect(prompt).toContain("the character's story");
+    expect(prompt).not.toMatch(/\bthe hero\b/i);
+  });
+
   it('returns 500 when request parsing fails', async () => {
     const response = await POST(makeRequest('invalid json'));
     const data = await response.json();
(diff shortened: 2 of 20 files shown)


jerseycheese:
Added one commit beyond the sweep's finding list, at Jack's request.

`CONTEXT.md` rules "hero" out of the vocabulary ("**Character**: _Avoid_: player character, PC, hero, avatar, protagonist"), and the ending path was ignoring that in five places. Three of them sit inside the prompt text that steers the image model, so a triumphant ending in a slasher world was asking for "the hero standing tall against a bright sky".

The tone guidance now interpolates the character's name instead of a generic noun, and the unnamed fallback reads "the character". `hero banner` in the layout requirements is untouched, since that's the web term for a wide image band.

Two tests cover it, and both fail against the previous commit: 2 failed / 7 passed without the fix, 9 passed with it.

Not fixed here, and worth its own call: the same function defaults `genre` to `'fantasy'` and the world name to `'Unknown Realm'`/`'the realm'`. Same genre-agnosticism problem, different word.

